PR #5, opened 2021-12-14
Initial Telescope extension

kabouzeid:
Been using this for myself. Let me know what you think :)


mickael-menu:
Thanks for your contribution!

I tried to add your fork as a Vim plugin with:

```
Plug 'kabouzeid/zk-nvim'
```

But then running `require("telescope").load_extension("zk")` I get the following error:

Screenshot 2021-12-14 at 20 01 48

Do you know what I might be missing in my config?

kabouzeid:
It's currently only on the telescope branch.

`Plug 'kabouzeid/zk-nvim', { 'branch': 'telescope' }`

kabouzeid:
Alternatively you can check out the repo locally and then use something like `Plug '/my/plugin/path'`

strange:
Great to see some activity here! Feels like `zk` is gaining momentum and plugins for the most popular editors is imperative to its success.

I started looking into creating something similar, alas Christmas is our busiest time of the year, and I haven't found the time.

I'll take this for a spin during my lunch break!

This is how I designed a theoretical plugin on a napkin:

- A command to open a telescope picker that can be filtered (title and tags is probably sufficient for a first version). I can press `<C-n>` to create a new note with the current filter as title (tags excluded). I can press `<CR>` to edit a selected file and I can navigate between files with `<C-j>` and `<C-k>`. `<C-c>` clears the filter state. If I run the command to open the picker again it remembers the last filter state (optional).
- Preview-support is nice, but I often keep so many terminals open, that even on my 34" monitor, I rarely have space for more than 80 columns.
- Basic LSP integration (complete tags, titles/paths when creating links etc).
- A command that opens up a picker with all files that are related to the file open in the current buffer.

kabouzeid:
Just a heads up, you need to build the latest (unreleased) version of zk from git for this to work, ie it doesn't work with zk 0.8.0.

Nothing needs to be done in this plugin for LSP integration, link/tag autocompletion etc already work out of the box with zk and lspconfig.

I've also included a "backlinks" picker which lists all notes that reference the currently open note.

mickael-menu:
@kabouzeid Thanks, I was able to run your plugin, it's working great!

I'll wait to release the version 0.9 (after fixing https://github.com/mickael-menu/zk/issues/118#issuecomment-992864682) before merging your PR, to list a proper version requirement. I guess I'll do this this week-end.

There's also some initial work that has been done by @smithbm2316, @eric-hansen and @cljoly in some alternate branches or forks. I'm not sure what's the state of this, and if we need to merge your work together. Would you mind chiming in?

cljoly:
> There's also some initial work that has been done by @smithbm2316, @eric-hansen and @cljoly in some alternate branches or forks. I'm not sure what's the state of this, and if we need to merge your work together. Would you mind chiming in?
> 
> >

I made a telescope picker by calling the `zk list` command (it was before you added `list` to the lsp) and I think the solution of this PR is much better.

smithbm2316:
> I made a telescope picker by calling the `zk list` command (it was before you added `list` to the lsp) and I think the solution of this PR is much better.

I've unfortunately been very caught up with work the last couple of weeks since I pushed some initial ideas/code to my branch a few weeks ago. We're at the end of a project lifecycle at the moment, so it's been a bit busier than I was hoping for me. @cljoly The `require('zk').list()` function that I wrote was meant to be an API wrapper around `zk list` so that people who don't use Telescope can still pump results from the `zk list` command into their favorite fuzzy finder in Neovim more easily. I hadn't gotten around to working on the Telescope extension that I was planning on at the time. I wanted to work on the API wrapper functions first before working on the Telescope-specific stuff.

@mickael-menu I just saw that you added the new LSP support for `zk list` which is *so EXCITING!!* I had written that `require('zk').list()` API wrapper before that was added, and the LSP method is most definitely the better way to go! I'm hoping to do some more work this weekend now that things have calmed down at work for me a bit to clean up that branch and get it ready for PR/review. Especially now that we have the `zk list` LSP command!

@kabouzeid thanks for submitting this, I can do a code review sometime this weekend if @mickael-menu or you would like me to before merging. I'll be excited to try out this branch tomorrow for myself 😄

lua/telescope/_extensions/zk.lua
```diff
@@ -0,0 +1,135 @@
...
+
+local function request_notes(bufnr, options, cb)
+  vim.lsp.buf_request(bufnr, "workspace/executeCommand", {
+    command = "zk.list",
+    arguments = {
+      vim.api.nvim_buf_get_name(bufnr),
+      vim.tbl_extend("error", { select = { "title", "absPath", "rawContent" } }, options or {}),
+    },
+  }, function(err, res)
+    assert(not err, err)
```
cljoly, line 19:
```suggestion
    assert(not err, tostring(err))
```
in my experience, `err` was a table and needed to be converted to a string.

lua/telescope/_extensions/zk.lua
```diff
@@ -0,0 +1,135 @@
...
+    end
+  end)
+end
+
+local function request_tags(bufnr, cb)
+  vim.lsp.buf_request(bufnr, "workspace/executeCommand", {
+    command = "zk.tag.list",
+    arguments = { vim.api.nvim_buf_get_name(bufnr) },
+  }, function(err, res)
+    assert(not err, err)
```
cljoly, line 31:
```suggestion
    assert(not err, tostring(err))
```
same as https://github.com/mickael-menu/zk-nvim/pull/5/files#r771732662 here

cljoly:
> the `require('zk').list()` function that I wrote was meant to be an API wrapper around `zk list`

@smithbm2316 I guess the two could be complementary: on one hand, the tightly integrated LSP call to `list` with telescope and, on the other hand, the more free form call to `zk list` with a template, that might be easier to integrate with other finders.

lua/telescope/_extensions/zk.lua
```diff
@@ -0,0 +1,135 @@
...
+            table.insert(selection, entry.value.name)
+          end)
+
+          if vim.tbl_isempty(selection) then
+            selection = { action_state.get_selected_entry().value.name }
+          end
+
+          actions.close(prompt_bufnr)
+
+          _list_notes(opts, { tags = selection })
```
cljoly, line 121:
```suggestion
          _list_notes(opts, {prompt_title = "Zk Notes for tag(s) "..vim.inspect(selection) }, { tags = selection })
```

I think that `{tags = selection}` needs to be passed to LSP and so as a third argument. That’s the only way I could get the `tags` command to then display a list of notes that have the selected tag.

cljoly:
Thanks a lot for this plugin @kabouzeid!
I’ve been able to get all the three commands to work and it’s great!

lua/telescope/_extensions/zk.lua
kabouzeid: Yes you're right, good catch! Must've overlooked this one during refactoring.

mickael-menu:
Thanks for chiming in guys! Looks like we can merge this PR after addressing the tags issue and then build from there.

I found another issue for could be addressed in another PR: the preview window doesn't automatically wrap the content, so if you don't format your Markdown you can't see most of the note content.

By the way @mhanberg also did some great work [in another repo](https://github.com/mhanberg/zk.nvim). He uses `fzf` instead of Telescope, but I think the LSP communication layer could be shared, as well as some utilities for example related to colors.

(@mhanberg Do you have any interest in participating to this repository directly, or you prefer maintaining your custom plugin?)

I opened a new discussion to talk about the architecture and features of this plugin: https://github.com/mickael-menu/zk-nvim/discussions/6

Let me know if there are other LSP commands missing in `zk`. I guess it could be interesting to have access to the config layer.

mhanberg:
> By the way @mhanberg also did some great work in another repo.

thanks!

> I think the LSP communication layer could be shared, as well as some utilities for example related to colors.

I agree

> (@mhanberg Do you have any interest in participating to this repository directly, or you prefer maintaining your custom plugin?)

I would rather unify on a single plugin, so I'm open to contributing PRs and collaborating on ideas in this repository.

kabouzeid:
@cljoly thanks for the review, you were right on all points :)

kabouzeid:
I've added some documentation and made a whole plugin out of it.

Most notably this doesn't use lspconfig to manage the language server, and will now only launch a single instance of the zk server. The notebook path can be optionally provided so that one can run the Telescope pickers and other commands without having a note open.

I've also added API wrappers for index, new, list, and tag.list, which should make it easy to write e.g. an FZF picker.

I appreciate your feedback :)

cljoly:
> The notebook path can be optionally provided so that one can run the Telescope pickers and other commands without having a note open.

Just a random idea, it does not look like you use the global variable `$ZK_NOTEBOOK_DIR`, maybe its value could be used as a default if no notebook path is provided? (disclaimer: I’ve looked into this really quickly, I hope to test tomorrow).

kabouzeid:
When no path is provided, the default is already the path of the current buffer.

I didn't know about this environment var.
In this case I think a good solution would be to only use the path of the current buffer as default if it is a note in a notebook. Otherwise use `$ZK_NOTEBOOK_DIR` as default. It's an easy change, I'll do it when I'm back home.

kabouzeid:
Actually no changes necessary. I just checked, and `zk` already automatically falls back to `$ZK_NOTEBOOK_DIR`, which is the correct behavior 🎉

mickael-menu:
Thanks for tackling the Lua wrapper API! I assume you took a look at the [architecture discussion](https://github.com/mickael-menu/zk-nvim/discussions/6)?

> Most notably this doesn't use lspconfig to manage the language server, and will now only launch a single instance of the zk server. The notebook path can be optionally provided so that one can run the Telescope pickers and other commands without having a note open.

That's an interesting take, is it to allow running the commands even if we are not inside a notebook?

I suggest we focus on merging your PR now to get the plugin quick-started from this great base and then build up more from there. Let me know when you're ready to merge.

All of this is very exciting!

lua/zk.lua
```diff
@@ -0,0 +1,22 @@
...
+M.cmd = require("zk.cmd")
+
+function M.setup(options)
+  config.options = vim.tbl_deep_extend("force", config.defaults, options or {})
+  if config.options.lsp.auto_attach.enabled then
+    util.setup_lsp_auto_attach()
+  end
+end
+
+vim.cmd("command! ZkIndex lua require('zk').cmd.index()")
```
smithbm2316, line 19: I think that we could potentially have a configuration option like `create_user_commands` that defaults to `true` so that users who don't want to have those can turn that off easily, and just wrap these `vim.cmd` calls in 
```lua
if config.options.create_user_commands then
  vim.cmd("command! ZkIndex lua require('zk').cmd.index()")
  vim.cmd("command! -nargs=? ZkNew lua require('zk').cmd.new(nil, { dir = [=[<args>]=]})")
end
```

lua/zk/util.lua
```diff
@@ -0,0 +1,36 @@
...
+    trigger = "FileType " .. table.concat(filetypes, ",")
+  else
+    trigger = "BufReadPost *"
+  end
+  vim.api.nvim_command(string.format("autocmd %s lua require'zk.util'.lsp_buf_auto_add(0)", trigger))
+end
+
+function M.is_notebook_path(path)
+  -- check that we got a match on the .zk root directory
+  return require("lspconfig.util").root_pattern(".zk")(path) ~= nil
```
smithbm2316, line 20: I'm sure it'll be quite obvious to most users, but we'll want to make sure that we note `nvim-lspconfig` as one of the project dependencies in the README

smithbm2316 (approved):
This looks great @kabouzeid and pretty much makes my branch #7 just a duplicate and less fleshed-out implementation. Let's go with yours and we can continue to build from here with what you've put together!

kabouzeid:
> Thanks for tackling the Lua wrapper API! I assume you took a look at the architecture discussion?

Yes :) Right now the Telescope "adapter" is loosely oriented on the discussion. At least for now, I'd keep it as is. One could easily copy [this code](https://github.com/kabouzeid/zk-nvim/blob/telescope/lua/telescope/_extensions/zk.lua) and replace the Telescope pickers with FZF pickers or vim.ui.select() calls.

> That's an interesting take, is it to allow running the commands even if we are not inside a notebook?

Yes, exactly. And as a bonus, when opening multiple notebooks, a single zk server instance will be shared between them. Overall this simplified the code, since now the zk client doesn't need to be attached to the buffer, and we don't have to deal with duplicated responses because multiple servers are running.

When the current buffer is not within a notebook and the user did not explicitly specify the notebook location, it will fall back to `$ZK_NOTEBOOK_DIR`, just like with the `zk` cli.
Meaning we can now use `:Telescope zk notes` (and every other command for that matter) from any buffer, and it'll do the right thing.

> I suggest we focus on merging your PR now to get the plugin quick-started from this great base and then build up more from there. Let me know when you're ready to merge.

I agree. Once this is merged we can tackle the rest with more focused PRs, this one has already gotten out of hand 😅

As you suggested https://github.com/mickael-menu/zk-nvim/pull/7#issuecomment-998246206, I've now moved the higher level commands from `zk.cmd` to `zk`. I think this will clear up confusion on whether to use the commands in `zk.api` or in `zk.cmd`

I'm quite satisfied with the experience of the plugin now. Feel free to merge.

smithbm2316:
If you'd like to @kabouzeid, it might be nice to just grab lines 226-237 from [this file](https://github.com/mickael-menu/zk-nvim/blob/cc6e7b80cd6a8d4abe18f40b63d9e75e1b7e7513/lua/zk/init.lua#L226) so that @mickael-menu can get his jumping between note links in a file with this PR too 😄

cljoly:
I’ve tested commit `8825b93` and it worked flawlessly for me. FWIW I was using a custom path to zk, like so:
```lua
require("zk").setup{
  lsp = {
    config = {
      cmd = { "/home/cj/ghq/github.com/mickael-menu/zk/zk", "lsp" },
      name = "zk",
      -- init_options = ...
      -- on_attach = ...
      -- etc, see `:h vim.lsp.start_client()`
    },
  },
}

```

lua/zk/util.lua
kabouzeid: Agreed! :) It's already in the README though, you must've overlooked it.

lua/zk.lua
kabouzeid: Good point, will do!

kabouzeid:
> If you'd like to @kabouzeid, it might be nice to just grab lines 226-237 from [this file](https://github.com/mickael-menu/zk-nvim/blob/cc6e7b80cd6a8d4abe18f40b63d9e75e1b7e7513/lua/zk/init.lua#L226) so that @mickael-menu can get his jumping between note links in a file with this PR too smile

Hmm, I'm not quite sure about this one. I might have another language server attached which also provides diagnostics (think eg writegood or proselint via null-ls or efm). In that case this would not jump to the next link but simply to the next diagnostic hint.

Also, if we add this, then to be consistent it would also make sense to add
```lua
function preview_link()
  vim.lsp.buf.hover()
end
```
and so on, but it seems a bit redundant to me.

cljoly:
> > If you'd like to @kabouzeid, it might be nice to just grab lines 226-237 from [this file](https://github.com/mickael-menu/zk-nvim/blob/cc6e7b80cd6a8d4abe18f40b63d9e75e1b7e7513/lua/zk/init.lua#L226) so that @mickael-menu can get his jumping between note links in a file with this PR too smile
> 
> Hmm, I'm not quite sure about this one. I might have another language server attached which also provides diagnostics (think eg writegood or proselint via null-ls or efm). In that case this would not jump to the next link but simply to the next diagnostic hint.
> 
> Also, if we add this, then to be consistent it would also make sense to add
> 
> ```lua
> function preview_link()
>   vim.lsp.buf.hover()
> end
> ```
> 
> and so on, but it seems a bit redundant to me.



I agree, plus at this point, it’s probably better to merge this PR and address those things in an other PR.

kabouzeid:
Apparently it's possbile to filter by namespace in goto_next(), and we can get the zk namespace via `h: vim.diagnostic.get_namespace()` and `h: vim.lsp.diagnostic.get_namespace()`. The diagnostic severity then also needs to be user configurable, since it's configurable in the zk config. This should be in it's own PR though.

mickael-menu:
It's actually small enough that I can bind a key directly to this for now, thanks for the idea @smithbm2316 

Seems you're all happy about the current state, let's merge! Thank you @kabouzeid

smithbm2316:
> > If you'd like to @kabouzeid, it might be nice to just grab lines 226-237 from [this file](https://github.com/mickael-menu/zk-nvim/blob/cc6e7b80cd6a8d4abe18f40b63d9e75e1b7e7513/lua/zk/init.lua#L226) so that @mickael-menu can get his jumping between note links in a file with this PR too smile
> 
> Hmm, I'm not quite sure about this one. I might have another language server attached which also provides diagnostics (think eg writegood or proselint via null-ls or efm). In that case this would not jump to the next link but simply to the next diagnostic hint.
> 
> Also, if we add this, then to be consistent it would also make sense to add
> 
> ```lua
> function preview_link()
>   vim.lsp.buf.hover()
> end
> ```
> 
> and so on, but it seems a bit redundant to me.

Ah yeah I forgot that people might use multiple LSPs with `zk-nvim`. I do that for web development stuff, but just attach `zk` itself for my ZK notebook. I have a keybinding setup globally anytime I attach a LSP to a buffer to be able to jump between diagnostics myself. Glad it worked out for you to just copy and paste though @mickael-menu! I'm going to close #7 and we can continue discussion and further development on the excellent base you've already put together @kabouzeid 😄

PR #7, opened 2021-12-20
[WIP] Initial API functions

### LINKED ISSUE #4: Jump to the next link with <TAB>
One particular feature I like about Vimwiki is to jump to the next or previous link with `<(S-)Tab>`. This makes it very easy to browse a notebook.

Maybe it's something that could be added to `zk-nvim` as well.

smithbm2316:
I unfortunately had less time than I was hoping to this weekend. It ended up being a lot more hectic than I anticipated. However, I've done some refactoring of my initial commit to include new API functions for the new `zk.list` and `zk.tags.list` LSP commands, as well as `zk.next_link` and `zk.prev_link` functions to allow for easy binding to jump to the next and previous note links using Neovim's new `vim.diagnostic` API (this would close #4). Any and all feedback is welcome, and I can't promise that all of the functions work fully at the moment, but are rather initial ideas for how they might work. This currently is still a work in progress, but I think it at least should give an idea of the vision I had for the API wrappers, while #5 focuses on the tightly-integrated Telescope behavior. 

As soon as I have some time one evening this week I will continue to refine this, do some cleanup, add some `asserts` for the various configuration options for the `zk.list` and `zk.tags.list` commands once I have a chance to read through all of the options in more detail, and write some docs. @mickael-menu you should be able to just copy and paste the `zk.next_link` and `zk.prev_link` functions for now into your config if you want that Vimwiki-like jumping between next/previous links until this PR is finished and merged.


kabouzeid:
I've added an API wrapper yesterday which is now no longer telescope specific. 
@smithbm2316 could you check if this could also work for you?

There is `require("zk").api` which is a wrapper around the LSP commands, and `require("zk").cmd` which uses the api and implements some default behavior for `zk.new` and `zk.index`. The latter is meant to be directly used in a command or keybinding.

There are also examples in the [README](https://github.com/kabouzeid/zk-nvim/tree/telescope#readme).

mickael-menu:
> @mickael-menu you should be able to just copy and paste the `zk.next_link` and `zk.prev_link` functions for now into your config if you want that Vimwiki-like jumping between next/previous links until this PR is finished and merged.

Thanks, that's great!

Yeah I think we should focus on merging #5 first to avoid you working on two different wrapper implementations. Then we can iterate with more PRs.

> There is `require("zk").api` which is a wrapper around the LSP commands, and `require("zk").cmd` which uses the api and implements some default behavior for `zk.new` and `zk.index`. The latter is meant to be directly used in a command or keybinding.

The `api` namespace is perfect for the LSP commands. For the higher-level commands, maybe it would be more convenient to put them at the root level directly? e.g. `require("zk").new()`

smithbm2316:
Closing as we have decided to go with #5 as our initial base featureset

PR #8, opened 2021-12-20
option to skip creation of user commands

kabouzeid:
Just a quick follow up to address this https://github.com/mickael-menu/zk-nvim/pull/5#discussion_r772679758


smithbm2316 (approved):
:tada:

PR #12, opened 2021-12-22
Make :ZkNew and :ZkIndex consistent with :ZkList

kabouzeid:
## ZkNewLink

There is a new `ZkNewLink` command. Select some text in visual mode, and then execute
```viml
:'<,'>ZkNewLink
```
this will create a new note and use the last visual selection as the title while replacing the selection with a link to the new note.
You can also pass the usual options
```viml
:'<,'>ZkNewLink { edit = false, dir = "work", date = "yesterday" }
```

---

## Changes for ZkNew

```viml
:ZkNew daily
```
becomes

```viml
:ZkNew { dir = "daily" }
```
which is a bit more verbose, but is consistent with `:ZkList` and allows for something like

```viml
:ZkNew { dir = "daily", date = "yesterday" }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,12 +75,14 @@ However, this is always optional, and usually not necessary.
 ### VimL
 ```vim
 " Indexes the notebook
-:ZkIndex
+" params
+"   (optional) additional options, see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zkindex
+:ZkIndex [<options>]
 
 " Creates and opens a new note
 " params
-"   (optional) directory for the new note, relative to the notebook root
-:ZkNew [<directory>]
+"   (optional) additional options, see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zknew
+:ZkNew [<options>]
 
 " Opens a Telescope picker
 " params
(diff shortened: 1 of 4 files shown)


mickael-menu (approved):
Looking good, the `ZkNewLink` command is a bit redundant with the LSP code actions but it's much more flexible so still very useful.

Another similar use case is to extract the current visual selection as the body of the future note (as opposed to the title). This could be used to "refactor" a large note into smaller notes.

kabouzeid:
> Looking good, the ZkNewLink command is a bit redundant with the LSP code actions but it's much more flexible so still very useful.

Could you elaborate on this? I'm not sure I understand what you mean 🤔 

I have something in mind to make things more easily extendable by the user, similar to the `alias` feature of the CLI. I'll propose this soon in the architecture discussion to see what you all think about it. I think after that this plugin will be basically feature complete.

mickael-menu:
> Could you elaborate on this? I'm not sure I understand what you mean 🤔

See this discussion https://github.com/mickael-menu/zk/discussions/100#discussioncomment-1627700

> I have something in mind to make things more easily extendable by the user, similar to the `alias` feature of the CLI. I'll propose this soon in the architecture discussion to see what you all think about it.

Sounds good, looking forward to your idea!

kabouzeid:
Ah, I didn't know about that. A bit redundant indeed.
The code range action doesn't seem to work reliably on a buffer with unwritten changes, i.e. the range seems to be off then. Not sure if that's Neovim's or zk's fault though.

PR #14, opened 2021-12-25
custom commands, FZF, vim.ui.select, refactoring

kabouzeid:
I've already updated most parts of the README but I'm not done yet. I'll also add a vimdoc file.

I've reorganized some stuff to improve extensibility by the user. Please have a look at [Custom Commands](https://github.com/mickael-menu/zk-nvim/tree/user-commands#custom-commands) in the README.
The [builtin commands](https://github.com/mickael-menu/zk-nvim/blob/user-commands/lua/zk/commands/builtin.lua) are defined in the same way as the commands that the user can now define in the setup function.

There's now also support for FZF and vim.ui.selection, configurable via the setup function.

The builtin commands now also include `:ZkCd`, `:ZkMatch`, `:ZkNewFromTitle`, `:ZkNewFromSelection`, and more, see [commands](https://github.com/mickael-menu/zk-nvim/tree/user-commands#commands).

What do you think about this? Some feedback would be nice before I work on finishing the documentation. @mickael-menu @smithbm2316 @cljoly


README.md
```diff
@@ -54,7 +53,14 @@ require("zk").setup({
       enabled = true,
       filetypes = { "markdown" },
     },
+
+    -- can be "telescope", "fzf" or "select" (`vim.ui.select`)
+    picker = "telescope",
```
cljoly, line 58: I’m not entirely sure, but it might be wiser to default to what everyone has installed, i.e. `vim.ui.select`.

cljoly: Or we can detect if telescope is installed and if not fallback to `vim.ui.select`.

cljoly:
It appears the orphans command is now to be added manually and that’s fine with me.

lua/zk/config.lua
```diff
@@ -12,8 +11,9 @@ M.defaults = {
       filetypes = { "markdown" },
     },
   },
+  picker = "telescope"
```
mickael-menu, line 14: I would use the `select` default picker as it doesn't require any other plugin.

cljoly:
While testing the changes, it seems there is a performance issue (at least with the rest of my config) where nvim hangs using ~110% CPU. This also seem to cause highlighting to be disabled in the markdown file.

I’ll try to narrow down where that comes from…

README.md
```diff
@@ -87,97 +93,141 @@ However, this is always optional, and usually not necessary.
...
 ### Lua
 ```lua
 ---Indexes the notebook
 --
----@param path? string path to explicitly specify the notebook
 ---@param options table additional options
+---@param path? string path to explicitly specify the notebook
 ---@see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zkindex
-require("zk").index(path, options)
+require("zk.commands").index(options, path)
```
mickael-menu, line 148: Why did you move the main functions under a `zk.commands` module? I think keeping it at the root level would make it more convenient to setup key bindings and make it clearer what are the user high-level functions.

README.md
```diff
@@ -87,97 +93,141 @@ However, this is always optional, and usually not necessary.
...
+---@see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zklist
+require("zk.commands").match(options, path)
+
+---Opens a Telescope picker
+--
 ---@param options table additional options
+---@param path? string path to explicitly specify the notebook
 ---@see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zktaglist
-require("zk").tag.list(path, options)
+require("zk.commands").tags(options, path)
```
mickael-menu, line 206: I think keeping `zk.tag.list()` would be better as I will add more `zk tag` commands later on (like `rename` or `remove`).

cljoly:
This seem to have been introduced from the first commit `9b6447f17bc130490d89a71972138b8c16f4d0c3`

kabouzeid:
How can I reproduce this? Or during which action does this happen?

mickael-menu (commented):
Thanks, that's a great refactoring. I really like the ability to add custom commands.

I have some suggestions regarding the API, to simplify it a bit and make it more cohesive with the `zk` CLI.

First, I would rename the `zk.pickers` module into `zk.ui`, as we might create other type of UI elements and this is similar to `vim.ui`.

Then, from lower-level to user-level, we could have the following API/building blocks:

* `zk.ui.pick(items, format, callback, options)`
    * Generic item picker using the user-selected picker.
    * Will call the `format` callback function to generate a string title for each item.
    * Will call `callback` with the list of picked items.
    * `options` can be generic options (e.g. `title` for the picker title) or specific options for third-party pickers.
* `zk.ui.pick_notes(notes, callback, options)`
    * Present the given list of `notes` with `zk.ui.pick()`.
    * Will call `callback` with the list of selected notes.
    * This function requires that the note tables contain at least `title` and `absPath`.
* `zk.pick_notes(callback, options)`
    * User-level function to query `zk` and pick notes.
    * Calls `zk.api.list()` then `zk.ui.pick_notes()` with the result.
    * Will call `callback` with the list of selected notes.
    * The `options` table will be passed to `zk.api.list()`, the sub-table under the `picker` key will be passed to `zk.ui.pick_notes()`.
    * The returned notes will contain at least `title` and `absPath`, but maybe more if additional `select` options are provided by the user.
* `zk.edit(options)`
    * Calls `zk.pick_notes()` and then edit/open the selected notes.
    * I renamed this from `zk.notes()` to make it clearer that we open the notes for edition, and more importantly, to use the same command names as the `zk` CLI.
    * (Maybe?) We could also add an `interactive: false` option  to skip the note picker and open the notes right away. Basically the default behavior of `zk edit [filtering]`.


* `zk.ui.pick_tags(tags, callback, options)`
    * Present the given list of `tags` with `zk.ui.pick()`.
    * Will call `callback` with the list of selected tags.
    * This function requires that the tag tables contain at least `name`.
* `zk.pick_tags(callback, options)`
    * User-level function to query `zk` and pick tags.
    * Calls `zk.api.tag.list()` then `zk.ui.pick_tags()` with the result.
    * Will call `callback` with the list of selected tags.
    * The `options` table will be passed to `zk.api.tag.list()`, the sub-table under the `picker` key will be passed to `zk.ui.pick_tags()`.
    * The returned tags will contain at least `name`, but maybe more if additional `select` options are provided by the user.
* `zk.edit_from_tags(options)`
    * Calls `zk.pick_tags()` and then `zk.edit({tag = })` with the result.

With this architecture it will be very easy to add custom commands, for example one to select notes to delete or to link in the current buffer.

README.md
mickael-menu: Agreed

cljoly:
> How can I reproduce this? Or during which action does this happen?

I’m trying to get a small config that triggers the issue…

cljoly:
Ok, so the issue with syntax coloration appears as soon as I open any markdown file and as long as I have `require("zk").setup{}` in myconfig. I’ve tried passing different arguments to setup, without success.

README.md
pinpox: At this point might as well also look for fzf?

smithbm2316:
> I've already updated most parts of the README but I'm not done yet. I'll also add a vimdoc file.

We can look into adding a GitHub action that creates a Vimdoc from our markdown documentation, anytime we push changes to the README. I can look to make a PR for that, I've not been online the last few days with the holidays! I'll take a look at what you've pushed soon to give some feedback! Thanks so much for the hard work on this @kabouzeid it's very exciting 😁

README.md
cljoly: That’s a good point, I guess the default setting could try to guess which picker is install and fallback to `vim.ui.select`.

My only concern would be that it might significantly slow down nvim startup time, because we would be exploring the various paths multiple times (I realize there is some caching here, but still). That would be a reason to default to `vim.ui.select`, so that there is slowdown due to looking up for many things.

lua/zk.lua
```diff
@@ -1,115 +1,77 @@
...
----Creates a new note
---
----@param path? string path to explicitly specify the notebook
----@param options table additional options
----@see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zknew
-function M.new(path, options)
-  -- neovim does not yet support window/showDocument, therefore we handle options.edit locally
-  if options and options.edit then
-    options.edit = nil -- nil means true in this context
+  if not M.notebook_root(vim.api.nvim_buf_get_name(bufnr)) then
```
cljoly, line 27:
```suggestion
  if not util.notebook_root(vim.api.nvim_buf_get_name(bufnr)) then
```
paired with the following at the top of the file:
```lua
local util = require("zk.util")
```
It does not seem to introduce any kind of circular dependency.

I think that’s the cause of the slowdown I mentioned earlier. The autocommand was entering some kind of loop, due to the call to the `nil` field (`M.notebook_root`).

cljoly:
> How can I reproduce this? Or during which action does this happen?

It appears the problem was solved when I made the change described in https://github.com/mickael-menu/zk-nvim/pull/14#discussion_r775293056

README.md
kabouzeid: I'll measure if there is any slow down probing for telescope > fzf > vim.ui.select.

README.md
```diff
@@ -87,97 +93,141 @@ However, this is always optional, and usually not necessary.
 " Creates a new note and uses the last visual selection as the title while replacing the selection with a link to the new note
 " params
 "   (optional) additional options, see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zknew
-:ZkNewLink [<options>]
+:'<,'>ZkNewFromTitleSelection [<options>]
+
+" Creates a new note and uses the last visual selection as the content while replacing the selection with a link to the new note
+" params
+"   (optional) additional options, see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zknew
+:'<,'>ZkNewFromContentSelection [<options>]
```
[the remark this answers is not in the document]
kabouzeid: I'd do it like you proposed here, what do you think is wrong with it?

README.md
```diff
@@ -87,97 +93,141 @@ However, this is always optional, and usually not necessary.
...
-require("zk").new_link() -- this will use your last visual mode selection
+require("zk.commands").new({ dir = "daily" })
+require("zk.commands").notes({ createdAfter = "3 days ago", tags = { "work" } })
+require("zk.commands").new_from_title_selection() -- this will use your last visual mode selection
 ```
 
 As you can see, the `path` is optional, and can usually be omitted; see [Notebook Directory Discovery](#notebook-directory-discovery).
 
 ### Telescope
+It's not really necessary to use this interface, instead we recommend you to use the provided [commands](#commands).
```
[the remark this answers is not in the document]
kabouzeid: Yeah, at this point we could probably just remove it from the documentation to avoid confusion. There aren't really any pros in using the telescope API directly.

README.md
kabouzeid: For consistency. The is zk.list command became zk.notes, because that fits better when the other commands are called zk.backlinks etc. Having that, I think zk.tags fits better with zk.notes. For rename, remove we could add zk.tag_rename or zk.rename_tag.

kabouzeid:
> Thanks, that's a great refactoring. I really like the ability to add custom commands.
> 
> I have some suggestions regarding the API, to simplify it a bit and make it more cohesive with the `zk` CLI.
> 
> First, I would rename the `zk.pickers` module into `zk.ui`, as we might create other type of UI elements and this is similar to `vim.ui`.
> 
> Then, from lower-level to user-level, we could have the following API/building blocks:
> 
> * `zk.ui.pick(items, format, callback, options)`
>   
>   * Generic item picker using the user-selected picker.
>   * Will call the `format` callback function to generate a string title for each item.
>   * Will call `callback` with the list of picked items.
>   * `options` can be generic options (e.g. `title` for the picker title) or specific options for third-party pickers.
> * `zk.ui.pick_notes(notes, callback, options)`
>   
>   * Present the given list of `notes` with `zk.ui.pick()`.
>   * Will call `callback` with the list of selected notes.
>   * This function requires that the note tables contain at least `title` and `absPath`.
> * `zk.pick_notes(callback, options)`
>   
>   * User-level function to query `zk` and pick notes.
>   * Calls `zk.api.list()` then `zk.ui.pick_notes()` with the result.
>   * Will call `callback` with the list of selected notes.
>   * The `options` table will be passed to `zk.api.list()`, the sub-table under the `picker` key will be passed to `zk.ui.pick_notes()`.
>   * The returned notes will contain at least `title` and `absPath`, but maybe more if additional `select` options are provided by the user.
> * `zk.edit(options)`
>   
>   * Calls `zk.pick_notes()` and then edit/open the selected notes.
>   * I renamed this from `zk.notes()` to make it clearer that we open the notes for edition, and more importantly, to use the same command names as the `zk` CLI.
>   * (Maybe?) We could also add an `interactive: false` option  to skip the note picker and open the notes right away. Basically the default behavior of `zk edit [filtering]`.
> * `zk.ui.pick_tags(tags, callback, options)`
>   
>   * Present the given list of `tags` with `zk.ui.pick()`.
>   * Will call `callback` with the list of selected tags.
>   * This function requires that the tag tables contain at least `name`.
> * `zk.pick_tags(callback, options)`
>   
>   * User-level function to query `zk` and pick tags.
>   * Calls `zk.api.tag.list()` then `zk.ui.pick_tags()` with the result.
>   * Will call `callback` with the list of selected tags.
>   * The `options` table will be passed to `zk.api.tag.list()`, the sub-table under the `picker` key will be passed to `zk.ui.pick_tags()`.
>   * The returned tags will contain at least `name`, but maybe more if additional `select` options are provided by the user.
> * `zk.edit_from_tags(options)`
>   
>   * Calls `zk.pick_tags()` and then `zk.edit({tag = })` with the result.
> 
> With this architecture it will be very easy to add custom commands, for example one to select notes to delete or to link in the current buffer.
> 
> 

I really like this architecture that you proposed, will work on it right now 👍🏽

README.md
mickael-menu: Okay, that makes sense. So I guess we can consider that the "edit/open" action is the default for every functions named as nouns?

README.md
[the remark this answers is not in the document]
mickael-menu: Nothing actually, I just wondered if there was an easier way to handle that from a Lua API, I'm not very familiar with what's available.

lua/zk.lua
kabouzeid: thanks, this is fixed now

kabouzeid:
I've added `zk.ui.pick_notes()` and `zk.ui.pick_tags()` but *not* `zk.ui.pick()` because that would've been too general.
I'm not a huge fan of the `zk.edit_from_tags()` name, but I also can't think of something better right now.
Documentation is not updated yet.

Apart from that, I'm quite happy with the current state and it's ready to review again :)

mickael-menu:
@kabouzeid In the end I think you were right with your initial naming using `notes()` and `tags()` to go with `backlinks()` and such. I guess since we are in an editor, the default action for APIs named as nouns should be "edit" and can be implicit.

mickael-menu:
Apart from that it's looking good! Do you have any other work in mind for this PR, besides updating the README?

cljoly:
I’ve tested again all the exposed command and direct calls to `Telescope zk …`, all seem to work in my config. Thanks for the refactoring @kabouzeid !

smithbm2316 (commented):
@kabouzeid I finally got the chance to sit down, play with this branch in my local Neovim setup, read through your code and the documentation. I've only added some small README notes just so you don't forget to make the tweaks, because this all looks quite good to me and has been really easy to use! I think the documentation and features you've written all work really well and are quite easy to use. I will work on the aforementioned suggestion of adding a Github action to automate the creation of a Vimdoc from this Markdown documentation once we merge this PR!

kabouzeid:
I've once again slightly changed how to define custom commands. I just want to get this right before we release, because the plugin is gaining momentum and I want to avoid major breaking changes in the near future if possible.

See https://github.com/mickael-menu/zk-nvim/tree/user-commands#custom-commands

Basically

```lua
require("zk").setup({
  commands = {
    foo_bar = {
      command = "ZkFooBar",
      fn = function(options)
        -- Do something
      end,
    }
  }
})
```

becomes

```lua
require("zk").setup()
require("zk.commands").foo_bar = {
  command = "ZkFooBar",
  fn = function(options)
    -- Do something
  end,
}
```

which is a bit more flexible.

eric-hansen:
@kabouzeid instead of it being a dynamic property why not just have a `require('zk.commands').add(<cmd name>, <setup func>)`??

So given your exmaple:

```lisp
require('zk.commands').add('ZkFooBar', function(options) end)
```

kabouzeid:
We're missing the name of the Lua function then.

It should also be available as a Lua function via `require("zk.commands").foo_bar(options)`

eric-hansen:
@kabouzeid is there a reason why you couldn't have `require('zk.commands').get(<cmd>[, options])`??

I.e.:

```
require('zk.commands').add('ZkFooBar', function(options) end)

local cmd = require('zk.commands').get('ZkFooBar', { ...opts })
```

kabouzeid:
You have a point. I'd prefer.

```lua
require('zk.commands').add('ZkFooBar', function(options) end)

require('zk.commands').get('ZkFooBar')(options)

-- or even

require('zk.commands').ZkFooBar(options)
```

eric-hansen:
Either or.  I'd prefer the 1st way personally.  Doing dynamic properties is scary since it can hinder future API development or cause unnecessary confusion.  Main reason why I'm advocating against it.

kabouzeid:
I'd usually agree, but in this specific case it shouldn't hinder API development in the future since vim user command must start with a capital letter, while any API function we might add in the future will always start with a lower case letter.

I'd even prefer

```lua
require('zk.commands').ZkFooBar = function(options) end
```

for setting, but there still needs to be a way to pass `range_only = true`, which is then no longer possible.

For consistency it's probably best to have either add and get, or both operations dynamic. Since dynamic add is not expressive enough, I think I agree with having plain old add and get functions.

kabouzeid:
```lua
require('zk.commands').add('ZkFooBar', function(options) end)

require('zk.commands').get('ZkFooBar')(options)
```

and for range only commands
```lua
require('zk.commands').add('ZkFooBar', function(options) end, { range_only = true })

require('zk.commands').get('ZkFooBar')(options)
```

kabouzeid:
As a bonus, this API would then be *very* similar to the new vim.api.nvim_add_user_command() API.

mickael-menu:
I agree, the last rendition with `add/get` seems safer and more flexible. Good job!

kabouzeid:
done

mickael-menu (approved):
@kabouzeid I removed `zk.edit_from_tags()` as the name was not great and we have the `ZkTags` command we can use directly now. I figured it would probably not be much reusable outside `ZkTags`.

Feel free to merge if you're ready, thanks again!

lua/telescope/_extensions/zk.lua
```diff
@@ -9,7 +9,7 @@ end
 ---@param opts? table additional options for zk, telescope options, all optional and in one table
 ---@see https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#zktaglist
 local function show_tags(opts)
-  zk.edit_from_tags(opts, { picker = "telescope", telescope = opts })
+  require("zk.commands").get("ZkTags")(opts, { picker = "telescope", telescope = opts })
```
kabouzeid: This doesn't work, ZkTags only takes one argument and therefore ignores picker="telescope". I've removed it now.

PR #29, opened 2022-01-27
Change default note sort order

kabouzeid:
This or whatever the `zk` default sort order for `zk list` is makes more sense I think

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,12 +368,12 @@ local opts = { noremap=true, silent=false }
 vim.api.nvim_set_keymap("n", "<leader>zn", "<Cmd>ZkNew { title = vim.fn.input('Title: ') }<CR>", opts)
 
 -- Open notes.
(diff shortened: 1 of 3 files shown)


mickael-menu:
I think we might want to use the default sort value for `zk list`, because it changes depending on the given criteria. I also think it could be useful to be able to customize the default in the `zk` configuration file, per note group. For example you might want to sort by title for an archive folder but by date for a daily journal one.

kabouzeid:
yeah makes sense. how do I get the default `zk list` sort order?

mickael-menu:
If you remove the `sort = ` parameter I think that should use the default sort order?

kabouzeid:
You're right. I've removed the sort key now and instead set it in the default keymaps.

mickael-menu (approved):

PR #31, opened 2022-01-29
Don't swallow LSP errors in zk.api

kabouzeid:
Errors can for example happen when no `zk` notebook root is found. The `zk.api` level is the wrong place to consume these errors.


mickael-menu (approved):
Agreed 👍

PR #48, opened 2022-05-17
fzf: fix jumping to note with spaces in filename

zhengpd:
The space of `delimiter` in `fzf.lua` cause note with spaces in filename
not openable.

Picker: `fzf`

Example note (zsh.config - vi mode.md):

```
  # zsh.config - vi mode

  note body...
```


mickael-menu (approved):
Thank you!

Issue #44, opened 2022-03-17
's: expected string, got nil' on :ZkNotes


lukamanitta:
When `:ZkNotes` is executed, the following error occurs:

```
Error executing vim.schedule lua callback: vim/shared.lua:63: s: expected string, got nil
stack traceback:
	[C]: in function 'error'
	vim/shared.lua:608: in function 'validate'
	vim/shared.lua:63: in function 'gsplit'
	vim/shared.lua:128: in function 'split'
	...e/pack/packer/start/zk-nvim/lua/zk/pickers/telescope.lua:53: in function 'define_preview'
	...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:386: in function 'preview'
	...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:1066: in function 'refresh_previewer'
	...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:1015: in function 'set_selection'
	...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:1351: in function '_do_selection'
	...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:1313: in function 'cb'
	vim.lua:285: in function <vim.lua:285>
```

Once the error is dismissed, the Telescope picker is opened, with an empty results field. When something is typed, the field populates with results, and when the input prompt is cleared (last character is deleted), the above error occurs again, identically.

Neovim version 0.6.1

guzmonne:
I was getting the same error and I was able to track the bug to this line:

https://github.com/mickael-menu/zk-nvim/blob/d705faa82da042e7a75f7d244afd27c2d8f20830/lua/zk/lsp.lua#L10

For some reason the value of `config.options.lsp.config` is wrong. I don't know when it gets mutated though.

A workaround to make this work is updating the this value to `config.defaults.lsp.config`.

If I found the place where these table is mutated I'll submit a PR.

zhengpd:
I also got this and traced back to 

https://github.com/mickael-menu/zk-nvim/blob/d705faa82da042e7a75f7d244afd27c2d8f20830/lua/zk/pickers/telescope.lua#L53

The first note in my notes list is empty, which might cause `vim.split(entry.value.rawContent, "\n")` to be nil, raising the error.

The problem is gone after adding a line into the empty note file.

zhengpd:
Just submitted a fix https://github.com/mickael-menu/zk-nvim/pull/49

[...]

PR #49, opened 2022-05-18
fix(#44): 's: expected string, got nil' on :ZkNotes

zhengpd:
It's caused by preview of empty note content

Picker: telescope

Fix https://github.com/mickael-menu/zk-nvim/issues/44


mickael-menu (approved):
Looking good, thank you 👍

PR #54, opened 2022-06-07
Add tags picker to Telescope extension.

deforde:
Added a function to facilitate loading the tags picker as a Telescope extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,9 +473,9 @@ end)
 ```
 
 ## Telescope Plugin
-> Not recommended, instead just use the [:ZkNotes command](#built-in-commands).
+> Not recommended, instead just use the [:ZkNotes or :ZkTags commands](#built-in-commands).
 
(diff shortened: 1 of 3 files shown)


eric-hansen:
Can there be documentation as to what the options are for this?

deforde:
Good point - I added some text to the docs regarding the new functionality for the Telescope plugin.
Similarly to the `notes` picker, I have not explicitly documented the available options in the Telescope plugin section itself, as this is already documented elsewhere.
The available options are the same as those available to `zk.list`, as already reference in the docs.
Although I did have issues getting the `sort` option to work - but that seemed to affect the extant `:ZkNotes` command as well - I will investigate this more closely later and provide an update.

mickael-menu (approved):
Looking good, thank you @deforde!

@eric-hansen You are a collaborator on the repo, I think you should be able to approve and merge the pull requests.

PR #57, opened 2022-06-10
fix tag name selection in fzf

idevat:
In tag list in `fzf` the line format is `notes_count delimiter tag_name` (e. g. `3 sometag`). Incorrect regexp is used for parsing tag name from the selected line. It causes that the selected tag is empty and notes are not filtered by tag in the following notes list. This pull request should fix it.


mickael-menu:
Hi @idevat, thanks for submitting this PR.

How can I reproduce this error exactly? It works fine for me with my tags and `fzf`. Do you have a sample of a note with a problematic tag?

mickael-menu (approved):
I could reproduce the problem with your example.

Thank you for your fix and contribution @idevat!

PR #63, opened 2022-07-09
Lazily load note contents in Telescope

pkazmier:
What are your thoughts on this "optimization"?

Prior to this change, the `rawContents` of **all** notes were loaded via the LSP connection.  When the quantity of notes is large (30k), this results in more memory utilization as well as slower transfer times from the LSP server, which ultimately delays the opening of Telescope.

With this change, `rawContents` is no longer requested from the LSP server--only `title`, `absPath`, and `path` are loaded for each note. To generate previews, we use the "absPath" to open the file via the `buffer_previewer_maker` when Telescope needs to show a preview.

I tested the performance of this against 30,000 random notes of lorem ipsum with a median size of 1,464 bytes.  Memory utilization is much lower and the time to open the Telescope window is roughly twice as fast on my computer.

### DIFF
--- a/lua/zk/pickers/telescope.lua
+++ b/lua/zk/pickers/telescope.lua
@@ -10,7 +10,7 @@ local previewers = require("telescope.previewers")
 
 local M = {}
 
-M.note_picker_list_api_selection = { "title", "absPath", "path", "rawContent" }
+M.note_picker_list_api_selection = { "title", "absPath", "path" }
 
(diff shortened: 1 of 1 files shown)


mickael-menu (approved):
I think that makes a lot of sense, I added the `select` parameter for this reason as I noticed lagginess when using LSP auto-completion (pre `zk-nvim`).

I'll wait a few days before merging to let others the opportunity to comment first.

PR #84, opened 2022-11-09
fix selection of non ascii title

artur-shaik:
I made a conversion bytes based indexes of character position to char based before execute `zk.new` command. So now I can successfully extract titles from Cyrillic encoded text.

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -10,6 +10,7 @@ commands.add("ZkNewFromTitleSelection", function(options)
   local location = util.get_lsp_location_from_selection()
   local selected_text = util.get_text_in_range(location.range)
   assert(selected_text ~= nil, "No selected text")
+  location.range = util.convert_byteidx_to_charidx(location.range)
   zk.new(vim.tbl_extend("force", { insertLinkAtLocation = location, title = selected_text }, options or {}))
 end, { needs_selection = true })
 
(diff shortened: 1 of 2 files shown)


mickael-menu (approved):
Thank you!

UnkwUsr:
Sadly, this patch have a few bugs:

1. This kills previously worked function with selecting multiple lines (when first line becomes a title, and all trailing becomes a description). The problem is in parent note, where should be placed our new link.

   Example:

   ```
   example title
   description
   ```

   Visual select this two lines, make `ZkNewFromTitleSelection` and see messed result:

   ```
   [example title](usi2)description
   ```

2. This messes whole line (even if selected only part of it) if you do visual select till the latest symbol on the line

   Example:

   ```
   * asd my future title
         ^ cursor is there
   ```

   Type `v$` and make `ZkNewFromTitleSelection`. Result:

   ```
   [my future title](b7ku)my future title
   ```

   (Also see how heading text ("* asd ") disappeared)

   To avoid messing we can visual select with `v$h` (just jump to one symbol left), so I think problem is with newline character.

   P.S. before this commit `v$` also worked without messing

@mickael-menu @artur-shaik 
Should I create issue for this or will you revert commit and continue work on this pull req?

mickael-menu:
Thanks for spotting this @UnkwUsr. I will revert the PR for now, feel free to open a new PR to implement a fixed version @artur-shaik.

PR #95, opened 2022-12-30
add option to insert link from telescope

abdullatif-k:
Hello,

This lets you select notes in telescope and add them as links in the current buffer.

This is already part of my workflow personally, I thought it might be useful to include in the plugin. This works fine _but I don't think the PR should be accepted in its current form_. 

There might be a better way to configure the mapping. And as it stands, this inserts links as a markdown link. It should probably autodetect the desired format from a notebook's config. We could probably read the notebook config and use a pattern search to get the `link-format`. It's straightforward if it's either markdown or wiki links, but it's more involved if it's custom. Perhaps this is functionality that should be implemented in zk and exposed through the LSP?

### DIFF
--- a/lua/zk/config.lua
+++ b/lua/zk/config.lua
@@ -12,6 +12,9 @@ M.defaults = {
       filetypes = { "markdown" },
     },
   },
+  mappings = {
+    telescope_link_note = { {"i", "n"}, "<C-Y>" }
+  }
 }
 
 M.options = M.defaults -- not necessary, but better code completion
--- a/lua/zk/pickers/telescope.lua
+++ b/lua/zk/pickers/telescope.lua
@@ -7,6 +7,7 @@ local action_utils = require("telescope.actions.utils")
 local putils = require("telescope.previewers.utils")
 local entry_display = require("telescope.pickers.entry_display")
 local previewers = require("telescope.previewers")
+local config = require("zk.config")
 
 local M = {}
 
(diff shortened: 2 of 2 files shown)


mickael-menu:
Thanks, that would be a really useful addition.

By the way, you can get a link to a note formatted according to the config option using:

```sh
zk list note.md -q --format "{{link}}"
```

It can even generate relative links with `-W`:

```sh
zk list -W dir ../note.md -q --format "{{link}}"
```

I don't think `format` is exposed in the `zk.list` LSP command but I could add it if needed.

abdullatif-k:
Thanks for the tip! It does seem from [this line](https://github.com/mickael-menu/zk/blob/65061987431428f8a8b45a7511f902e401fdfd20/internal/adapter/lsp/cmd_list.go#L52) in the zk LSP server that format doesn't get used at all, as opposed to [the CLI version of the same command.](https://github.com/mickael-menu/zk/blob/65061987431428f8a8b45a7511f902e401fdfd20/internal/cli/cmd/list.go#L117) I can reopen this PR if `format` gets implemented in the LSP and refactor the code to use it. 

I would've made the changes in zk myself but I haven't written a line of Go in my life. I'm curious though, is there a reason that the LSP and the CLI  instantiate notes differently? Couldn't they share the exact implementation?

mickael-menu:
Okay, I'll let you know when something is available in the zk LSP. There is a pending PR that might be of interest for this as well: https://github.com/mickael-menu/zk/pull/284

The reason `format` is not available now is because it's used to format the output of the CLI command. But the LSP command returns a structured JSON, so a custom `format` doesn't make much sense in this context. 

> I'm curious though, is there a reason that the LSP and the CLI instantiate notes differently? Couldn't they share the exact implementation?

They do share the same implementation. The only exception is the CLI takes the `Note` model and produce a formatted output according to `format`, while the LSP command serializes the `Note` to JSON.

PR #97, opened 2023-01-26
Add support for 'zk.link' and some example linking commands

psanker:
Adding support for mickael-menu/zk#284

"zk.link" performs the file linking operation that `insertLinkAtLocation` does in "zk.new" but without creating a new note. This allows users to create new linking capabilities, such as using match to find related notes and creating a link from the found note.

This PR also adds two commands: 

- `ZkInsertLink`
- `ZkInsertLinkAtSelection`

Both use the note picker of choice to query all notes to find the desired note to create a link (solves #95)

### DIFF
--- a/lua/zk/util.lua
+++ b/lua/zk/util.lua
@@ -51,6 +51,29 @@ function M.get_lsp_location_from_selection()
   }
 end
 
+---Fix to correct cursor location
+--
+---When working on link insertion, it was discovered that there may be
+---an off-by-one error for single point locations in glsp. This function
+---corrects that error.
+---@param location table An LSP location object representing a single cell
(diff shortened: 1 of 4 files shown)


lua/zk/util.lua
```diff
@@ -51,22 +51,45 @@ function M.get_lsp_location_from_selection()
   }
 end
 
+---Fix to correct cursor location
```
mickael-menu, line 54: The nvim cursor is (1,0)-indexed, so lines start at 1. Might be what's causing the issue you noticed.

mickael-menu (approved):
That's a super useful addition, thanks again! The `matchSelected` option is a nice touch as well. 👍

Issue #104, opened 2023-02-16
Error executing vim.schedule Cannont insert link for `:ZkNewFromTitleSelection`  `:ZkNewFromContentSelection` 


mr-majkel:
Hi,

first of all thanks for the great tool that is `zk` and the neovim plugin! I am one of the poor users who are commited (forced?) to using zk on Windows. This issue is occurring for both Windows 10 and WSL with Ubuntu.

No matter if I use zk on native Windows or in WSL, I seem not to be able to use either of the linking commands coming with the plugin. However `:ZkNew` works, so the problem seems to be connected with inserting the link to the document. This is the error I'm getting (this one is from WSL run, but it differs from the Windows one only by how the path is formatted):

```
Error executing vim.schedule lua callback: ...cal/share/nvim/site/pack/packer/start/zk-nvim/lua/zk.lua:68: RPC[Error] code_name = InvalidRequest, message = "Cannot insert link in 'file:///mnt/c/Users/modzmi01/zk_notebook/index.md'"
stack traceback:
        [C]: in function 'assert'
        ...cal/share/nvim/site/pack/packer/start/zk-nvim/lua/zk.lua:68: in function 'handler'
        ...l/.local/nvim-linux64/share/nvim/runtime/lua/vim/lsp.lua:1383: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

I'm using nvim 0.8.3.

I'm configuring and attaching zk lsp through lspconfig, so I have turned off auto attach in the setup. Here are the bits of my config related to zk:

<details>

```lua
-- plugin config
require("zk").setup({
  -- can be "telescope", "fzf" or "select" (`vim.ui.select`)
  -- it's recommended to use "telescope" or "fzf"
  picker = "telescope",

  lsp = {
    -- `config` is passed to `vim.lsp.start_client(config)`
    config = {
      cmd = { "zk", "lsp" },
      name = "zk",
      -- on_attach = ...
      -- etc, see `:h vim.lsp.start_client()`
    },

    -- automatically attach buffers in a zk notebook that match the given filetypes
    auto_attach = {
      enabled = false,
      filetypes = { "markdown" },
    },
  },
})

```

```lua
--lsp setup
local zk_on_attach = function(client, bufnr)
  local function buf_set_option(...) vim.api.nvim_buf_set_option(bufnr, ...) end
  local function buf_set_keymap(...) vim.api.nvim_buf_set_keymap(bufnr, ...) end

[...]

PR #105, opened 2023-02-16
Use already attached externally setup lsp client if available

mr-majkel:
Hi @mickael-menu,

I have prepared a solution for #104 and more generally a situation when zk lsp is set up outside `zk-nvim` (e.g. through lsp config).

The reason for going this route is connected with the builtin `zk lsp` setup seems to be broken for Windows (that I know is not supported officially). I think that because of improper path handling the root_dir is not found. [`lsp config`](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/zk.lua#L7) is using their own util function to find a dir with .zk directory, that seems to do the trick.

I think that having an option to use externally started server is still a feature worth to have. In case multiple zk-lsp servers are started and successfully attached to any buffer, the first one found by name is used. If none are found, zk-nvim starts its own and it takes the precedence from then on.

### DIFF
--- a/lua/zk/lsp.lua
+++ b/lua/zk/lsp.lua
@@ -4,8 +4,34 @@ local client_id = nil
 
 local M = {}
 
+---Tries to find a client by name
+function M.external_client()
+  local client_name = config.options.lsp.config.name
+  if not client_name then
+    client_name = "zk"
+  end
+
+  local active_clients = vim.lsp.get_active_clients({name=client_name})
+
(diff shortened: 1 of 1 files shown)


mickael-menu (approved):
Thank you for taking the time to address this. It doesn't seem to break anything from what I could see. 👍

mickael-menu:
Oops, I thought I merged your PR. Sorry about that @mr-majkel.

PR #118, opened 2023-04-03
fix: escape special characters in path

dstanberry:
This is copied from `lua/lspconfig/util.lua` in nvim-lspconfig.
The root_pattern lookup will fail when the path contains certain special characters.

Fixes the issue discussed in [#2111](https://github.com/neovim/nvim-lspconfig/issues/2111)


mickael-menu (approved):
Thank you 👍

PR #149, opened 2023-10-21
Documentation for fzf_lua support

miikanissi:
I noticed the earlier pull [https://github.com/mickael-menu/zk-nvim/pull/138](https://github.com/mickael-menu/zk-nvim/pull/138) added support for fzf-lua as a picker but it is not documented yet. I only found out about it by checking the source.

Changed a couple lines on the docs and README to document it.

Hope this helps, thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ require("zk").setup()
 **The default configuration**
 ```lua
 require("zk").setup({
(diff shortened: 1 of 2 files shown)


mickael-menu (approved):
Thank you @miikanissi

PR #159, opened 2024-01-03
Return Root from `zk.cd` and add Dry Run option for scripting

mcDevnagh:


kabouzeid:
Should rather directly use the util functions when scripting. I would like to avoid this kind of feature creep.

PR #163, opened 2024-04-07
Add support for `mini.pick` as a picker

pkazmier:
With [mini](https://github.com/echasnovski/mini.nvim/tree/main) gaining in popularity, this commit adds support for [mini.pick](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-pick.md).

Here is a short screencast showing this new picker in action with Zk notes:

https://github.com/zk-org/zk-nvim/assets/747855/cf645482-c3eb-4713-8925-bb1501f54cb8

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Via [lazy.nvim](https://github.com/folke/lazy.nvim)
 }
 ```
 
-To get the best experience, it's recommended to also install either [Telescope](https://github.com/nvim-telescope/telescope.nvim) or [fzf](https://github.com/junegunn/fzf).
+To get the best experience, it's recommended to also install either [Telescope](https://github.com/nvim-telescope/telescope.nvim), [fzf](https://github.com/junegunn/fzf), or [mini.pick](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-pick.md)
 
 ## Setup
 
@@ -74,8 +74,8 @@ require("zk").setup()
 **The default configuration**
 ```lua
 require("zk").setup({
-  -- can be "telescope", "fzf", "fzf_lua" or "select" (`vim.ui.select`)
-  -- it's recommended to use "telescope", "fzf" or "fzf_lua"
+  -- can be "telescope", "fzf", "fzf_lua", "minipick", or "select" (`vim.ui.select`)
+  -- it's recommended to use "telescope", "fzf", "fzf_lua", or "minipick"
   picker = "select",
 
   lsp = {
--- a/lua/zk/pickers/minipick.lua
+++ b/lua/zk/pickers/minipick.lua
@@ -5,8 +5,9 @@ local minipick = require("mini.pick")
 M.note_picker_list_api_selection = { "title", "path", "absPath" }
 
 M.show_note_picker = function(notes, opts, cb)
-  notes = vim.tbl_map(function(n)
-    return { text = n.title, path = n.absPath, note = n }
+  notes = vim.tbl_map(function(note)
+    local title = note.title or note.path
+    return { text = title, path = note.absPath, note = note }
   end, notes)
 
   opts = opts or {}
--- a/lua/zk/pickers/minipick.lua
+++ b/lua/zk/pickers/minipick.lua
@@ -7,65 +7,69 @@ M.note_picker_list_api_selection = { "title", "path", "absPath" }
 M.show_note_picker = function(notes, opts, cb)
   notes = vim.tbl_map(function(note)
     local title = note.title or note.path
-    return { text = title, path = note.absPath, note = note }
+    return { text = title, path = note.absPath, value = note }
   end, notes)
-
-  opts = opts or {}
-  local minipick_opts = vim.tbl_deep_extend("force", {
-    source = {
-      items = notes,
-      name = opts.title,
-      choose = function(_)
-        return nil
-      end,
-    },
-    window = {
-      prompt_prefix = opts.title .. "> ",
-    },
-  }, opts.minipick or {})
-
-  local item = minipick.start(minipick_opts)
-  if item ~= nil then
-    cb(opts.multi_select and { item.note } or item.note)
-  end
+  H.item_picker(notes, opts, cb)
 end
 
 M.show_tag_picker = function(tags, opts, cb)
-  tags = vim.tbl_map(function(t)
-    local padded_cnt = H.ensure_text_width(t.note_count, opts.note_count_width or 4)
+  local width = H.max_note_count(tags)
+  tags = vim.tbl_map(function(tag)
+    local padded_count = H.pad(tag.note_count, width)
     return {
-      text = string.format("%s │ %s", padded_cnt, t.name),
-      tag = t,
+      text = string.format(" %s │ %s", padded_count, tag.name),
+      value = tag,
     }
   end, tags)
+  H.item_picker(tags, opts, cb)
+end
 
+H.item_picker = function(items, opts, cb)
   opts = opts or {}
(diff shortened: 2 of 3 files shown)


lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,73 @@
+local M = {}
+local H = {}
+local minipick = require("mini.pick")
+
+-- Picker from zk-nvim must implement this.
```
jurica, line 5: Tbh I don't really like that comment. The other picker implementations don't have "interface implementation" commentend either and I don't really see any added value in them.

lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,73 @@
...
+M.note_picker_list_api_selection = { "title", "path", "absPath" }
+
+-- Picker from zk-nvim must implement this.
```
jurica, line 8: Same as above.

lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,73 @@
...
+    },
+  }, opts.minipick or {})
+
+  local item = minipick.start(minipick_opts)
+  if item ~= nil then
+    cb(opts.multi_select and { item.note } or item.note)
+  end
+end
+
+-- Picker from zk-nvim must implement this.
```
jurica, line 34: Same as above.

jurica (commented):
Hi @pkazmier , that's very cool thank you for the contribution!
Apart from some minor nitpicks the code looks good to me. But I would be glad if @kabouzeid could also take a look.

@zk-org/maintainers how do we want to handle this contribution? As we're still in maintenance mode strictly taken we shouldn't merge it yet. On the other hand there isn't really a release process for zk-nvim currently.

tjex:
Although we've gotten through all the bugs for `zk-nvim`, I would consider us still in maintenance mode ([as per the readme](https://github.com/zk-org/zk-nvim?tab=readme-ov-file#news-we-are-in-maintenance-mode)) until we cut a new release of `zk`. This should realistically be quite soon, at which point we may want to look at starting with release versioning with `zk-nvim` as well. 

I would say we should continue holding off on all feature request PRs until the above has happened?

lua/zk/pickers/minipick.lua
pkazmier: Agreed. That was a comment I added to show the maintainer of mini how I integrated it here. I’ll remove it.

README.md
```diff
@@ -74,8 +74,8 @@ require("zk").setup()
 **The default configuration**
 ```lua
 require("zk").setup({
-  -- can be "telescope", "fzf", "fzf_lua" or "select" (`vim.ui.select`)
-  -- it's recommended to use "telescope", "fzf" or "fzf_lua"
+  -- can be "telescope", "fzf", "fzf_lua", "minipick", or "select" (`vim.ui.select`)
```
kabouzeid, line 77: what about naming it `mini.pick` or `mini` instead?

lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,70 @@
...
+  end, notes)
+
+  opts = opts or {}
+  local minipick_opts = vim.tbl_deep_extend("force", {
+    source = {
+      items = notes,
+      name = opts.title,
+      choose = function(_)
+        return nil
+      end,
```
kabouzeid, line 20: what does this do?

lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,70 @@
...
+    },
+    window = {
+      prompt_prefix = opts.title .. "> ",
+    },
+  }, opts.minipick or {})
+
+  local item = minipick.start(minipick_opts)
+  if item ~= nil then
+    cb(opts.multi_select and { item.tag } or item.tag)
+  end
```
kabouzeid, line 59: does mini not support multi selection?

lua/zk/pickers/minipick.lua
```diff
@@ -0,0 +1,70 @@
...
+end
+
+-- From mini.extras
+H.ensure_text_width = function(text, width)
+  local text_width = vim.fn.strchars(text)
+  if text_width <= width then
+    return string.rep(" ", width - text_width) .. text
+  end
+  return "…" .. vim.fn.strcharpart(text, text_width - width + 1, width - 1)
+end
```
kabouzeid, line 69: what is that for?

kabouzeid (commented):
Overall looks good to me and tried it out locally. Please clarify the code though.

lua/zk/pickers/minipick.lua
kabouzeid: I think it does. It's called "mark" in mini.pick.

README.md
pkazmier: Sure, I'll rename it to `mini.pick`.

lua/zk/pickers/minipick.lua
pkazmier: The `choose` function in `mini.pick` is used to take action on an item—such as opening a buffer to edit a file. Because `zk-nvim` already handles what to do with selected options, I set this to `nil`. The value selected is returned as a result of `MiniPick.start()`. I then pass that value to the provided `zk-nvim` callback to handle the "action" of what to do.

lua/zk/pickers/minipick.lua
pkazmier: It does support multi-selection, but the `MiniPick.start()` method only returns a single note even if multiple notes were selected. Similar to the above comment about `choose`,`mini.pick` does have a `choose_marked` to take action on multiple selected items. For the same reasons above, I want to delegate to the user supplied callback function, but I don't know how to get access to the ... [scratch that last sentence, as I was typing this comment, I realized I should just be invoking the user-supplied callback in the `choose` functions. Let me play around with this and will update this weekend. Thanks for asking the question].

lua/zk/pickers/minipick.lua
kabouzeid: Why do you need to specify it then? The default is already `choose = nil`. Does that not work for some reason?

lua/zk/pickers/minipick.lua
kabouzeid: why only 4 digits? if you really want to limit this number for some reason, I would rather use ">9999" instead of "…999"

lua/zk/pickers/minipick.lua
pkazmier: There is a difference between a `nil` value for `choose` and a function that returns `nil`. The former invokes the default action from `mini.pick` while the latter tells `mini.pick` to close the picker without executing an action. But as I mentioned in further comment, I'll be refactoring this.

lua/zk/pickers/minipick.lua
pkazmier: When I refactor this later this weekend, there will be no truncation as I'll simply compute the max width required, so the count field will never truncated, so that will eliminate the need to make any decision here.

README.md
pkazmier: I recall why I named it `minipick`. Because that option is just concatenated to `"zk.pickers"` and then required. So, if we rename that, then we need to rename the file as well, but I don't think you can have dots in the filename, right? And I prefer to name it "minipick" over "mini" as many users only use a subset of the mini umbrella.

pkazmier:
Refactored the code to address all of your comments: multi-select, padding, etc ...

kabouzeid (approved):

kabouzeid:
thanks for the contribution!

pkazmier:
Thank you!

PR #166, opened 2024-05-24
fix(doc): add vim modeline to properly show the help

tonyxwz:
Without this modeline, when I run `:help zk-zk-nvim`, the filetype is
txt (text), not help.

### DIFF
--- a/doc/zk.txt
+++ b/doc/zk.txt
@@ -490,3 +490,5 @@ It's possible (but unnecessary) to also load the notes and tags pickers as a tel
     :Telescope zk tags created=today
 <
 
+==============================================================================
+ vim:tw=78:ts=8:ft=help:norl:fdm=marker:cole=2:


wzxxing:
**Before**, filetype=text (my spell checker is on too)

Screenshot 2024-05-24 at 22 10 56

**After**

Screenshot 2024-05-24 at 22 09 17

tjex:
Hmm, can't reproduce on my end. Is it just the spell checker which is leading you to the conclusion?
i.e. does `lua print(vim.bo.filetype)` return `text` or `help`?

I take it the spell checker doesn't activate on other help files?
And you're fully up to date with neovim / plugin versions?

That all being said, it's good practice to include the modeline anyway, and can't do any harm @kabouzeid ?

kabouzeid (approved):

tonyxwz:
This is how Neovim detects help files


[<https://github.com/neovim/neovim/blob/93c55c238f4c1088da4dc6ec80103eb3ef4085d2/runtime/lua/vim/filetype/detect.lua#L1561-L1563>
](https://github.com/neovim/neovim/blob/93c55c238f4c1088da4dc6ec80103eb3ef4085d2/runtime/lua/vim/filetype/detect.lua#L1557-L1565)

PR #171, opened 2024-06-22
Zk Grep

shfattig:
Use Telescope's "jobs" to perform live asynchronous calls to `zk list -m` to provide grep-like searching of note bodies

### DIFF
--- a/lua/zk.lua
+++ b/lua/zk.lua
@@ -100,8 +100,8 @@ end
 function M.pick_notes(options, picker_options, cb)
   options = vim.tbl_extend("force", { select = ui.get_pick_notes_list_api_selection(picker_options) }, options or {})
 
-  if options["grep"] ~= nil then
-    picker_options["grep"] = true
+  if options.grep then
+    picker_options.grep = options.grep
     ui.pick_notes({}, picker_options, cb)
   else
     api.list(options.notebook_path, options, function(err, notes)
--- a/lua/zk.lua
+++ b/lua/zk.lua
@@ -99,10 +99,16 @@ end
 ---@see zk.ui.pick_notes
 function M.pick_notes(options, picker_options, cb)
   options = vim.tbl_extend("force", { select = ui.get_pick_notes_list_api_selection(picker_options) }, options or {})
-  api.list(options.notebook_path, options, function(err, notes)
-    assert(not err, tostring(err))
-    ui.pick_notes(notes, picker_options, cb)
-  end)
+
+  if options["grep"] ~= nil then
+    picker_options["grep"] = true
+    ui.pick_notes({}, picker_options, cb)
+  else
+    api.list(options.notebook_path, options, function(err, notes)
+      assert(not err, tostring(err))
+      ui.pick_notes(notes, picker_options, cb)
+    end)
+  end
 end
 
 ---Opens a tags picker, and calls the callback with the selection
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -61,6 +61,13 @@ commands.add("ZkNotes", function(options)
   zk.edit(options, { title = "Zk Notes" })
 end)
 
+commands.add("ZkGrep", function(options)
+  -- Insert an option to grep the notes
+  -- TODO: insert rather than overwrite
+  options = {grep=true}
+  zk.edit(options, { title = "Zk Grep" })
+end)
+
 commands.add("ZkBacklinks", function(options)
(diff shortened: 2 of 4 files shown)


shfattig:
To test, run `:ZkGrep`

shfattig:
closes #127

shfattig:
A few things to clean up, but should be functional right now. Additionally, as mentioned in #127 , `zk list -m` does not provide line numbers of matches, but doing so would greatly enhance the match preview as well as the opening of the file. Also, `zk list -m` seems to behave differently from `zk edit -i`, matching less frequently and omitting expected matches for some reason

tjex:
Tested by replacing the hard coded notebook root with my own directory. I can execute the command and grep :)
Great!

I'm going to have a look at how telescope grepper does the highlighting for the returned grep result in the previewer and see how trivial it is to implement. Thanks for this pr! 🙌

tjex:
> A few things to clean up, but should be functional right now. Additionally, as mentioned in #127 , `zk list -m` does not provide line numbers of matches, but doing so would greatly enhance the match preview as well as the opening of the file. Also, `zk list -m` seems to behave differently from `zk list -i`, matching less frequently and omitting expected matches for some reason
> 
> edit: compare to `zk list -i` (although `zk edit -i` seems to give model behavior as well)

It would seem that `zk list -i` only searches through note titles? This is why it's returning less results than `zk list -m <term>`.

Jumping to the line number would be great. We should be able to hook into the LSP implementation of this from `zk`, which does get used for `ZkInsertLinkAtSelection` among other commands (also `:ZkMatch`)

shfattig:
> Tested by replacing the hard coded notebook root with my own directory. I can execute the command and grep :) Great!
> 
> I'm going to have a look at how telescope grepper does the highlighting for the returned grep result in the previewer and see how trivial it is to implement. Thanks for this pr! 🙌

You're welcome! Yeah it should be fairly trivial as long as we can get match line numbers. I'll work on some of the cleanup (like removing that hard coded path) and attempt a call to the lua api instead

lingling9000:
Any updates on this feature? I would love to have this!

shfattig:
> Any updates on this feature? I would love to have this!

Haha thanks for the poke. I'm looking at using the zk lua api instead of an external shell command for the grepping. It doesn't look like Telescope has an existing finder for this, so I'll have to figure out how to get there.

I did find that changing the matching strategy to "re" (regex) does give better results (rather than waiting until the end of a word). It also gives more intuitive usage haha, given that this _is_ a **grep** feature after all...) - what do you think?

shfattig:
Other than that, maybe we should outline what the minimum requirements are to get this merged in. I might argue that it's functional as is and ready for public release, as long as code style is fine and docs are updated appropriately (I can go do that, but you might have to help me find the right places)

shfattig:
@tjex can you get us line numbers on `zk list -m`?

tjex:
I have a bit too much to tackle at the moment both here with zk (mainly new docs site, another active PR and some aging bugs) and my study / job hunting commitments to be able to deliver something like that in any respectable amount of time :( 

I would personally be fine with merging without the jump-to-line feature. You could also leave a comment where that step should be implemented, so that it can be addressed later. Better to have the cake, even if the cherry on top is missing for the moment imo.

Another approach could be to grab the user input search query, and after entering the file, executing a regular buffer neovim search, using the user input as the search term?

tjex:
@shfattig Maybe there is something to take from this discussion on zk-org/zk. 

https://github.com/zk-org/zk/discussions/330

This person was also looking to integrate jump to line from grep / fzf. Perhaps it works for your contribution as well.

shfattig:
> @shfattig Maybe there is something to take from this discussion on zk-org/zk. 
> 
> https://github.com/zk-org/zk/discussions/330
> 
> This person was also looking to integrate jump to line from grep / fzf. Perhaps it works for your contribution as well.

Yeah, interesting. I can explore that more. I still think this feature is worth merging as is, and getting it in front of people might help to refine the behavior we're looking for. I'll be able to take some time for this in a couple weeks

tjex:
Just realised that this implementation only works with using telescope as the picker. This wasn't clear in the code somehow! Like the other `:Zk...` commands, it should be picker agnostic. I'll have a look into this too. 

The grepper in `fzf-lua` manages to get line numbers returned too. Maybe some hints for how to get line numbers from `zk list -m` as a result.

tjex:
I'm having a little look at it too, cleaned up a few things just now.

PR #172, opened 2024-06-22
enforce current stylua style

tjex:
During #171 I noticed that the call_parenthesis stylua setting had not been set in `stylua.toml`. 

Updated the stylua.toml file accordingly in keeping with the current formatting in the project and formatted the whole project.

I added a line length of 110 as a sane default (and personal preference). 
It's fairly arguable to have some hard setting for line length to catch future contributions with line wrap on?..


tjex:
err. I guess I just found out by accident that I can push directly to this repo!.. Because I def didn't merge this myself. I guess GitHub resolves PRs if the commit is pushed to the same branch which is being merged into..?

Anyway. If this change was not ok, then ping me back here and I'll revert it.

PR #173, opened 2024-06-23
fix(setup): not really detecting external clients

hoofcushion:
Two Problem:

 - `tbl == {}` is not the right way to tell if a table is empty in Lua, which makes the `external_client` meaningless. 

 - If two client starts in the same event loop, it will start `zk` twice. e.g. when user try to use filetype autocmd to start `zk` and `zk-nvim` together.

Two Solution:

 - Use `next(tbl)==nil` or `vim.tbl_isempty(tbl)` instead.
 
 - Use `vim.schedule` to prevent multiple client starts in the same event loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ To get the best experience, it's recommended to also install either [Telescope](
 > If you have the [zk cli](https://github.com/zk-org/zk) installed, then you do not need to install the `zk lsp` via the regular neovim paradigm (Mason, nvim-lsp, etc).
 > This is because `zk` has our lsp bundled with it, which `zk-nvim` hooks into.
 > Additionally, `zk-nvim` will setup and start the LSP server for you, do *not* call `require("lspconfig").zk.setup()`.
+> If you already had `require("lspconfig").zk.setup()` in use, please set `opt.lsp.auto_attach` to false `zk-nvim` will find the existed instance of `zk` lsp client automatically.
 
 ```lua
 require("zk").setup()
--- a/lua/zk.lua
+++ b/lua/zk.lua
@@ -21,9 +21,9 @@ end
 
 ---Automatically called via an |autocmd| if lsp.auto_attach is enabled.
 --
----@param bufnr number
+---@param bufnr integer
 function M._lsp_buf_auto_add(bufnr)
-  if vim.api.nvim_buf_get_option(bufnr, "buftype") == "nofile" then
+  if vim.bo[bufnr].buftype == "nofile" then
     return
   end
 
--- a/lua/zk/api.lua
+++ b/lua/zk/api.lua
@@ -15,7 +15,6 @@ local function execute_command(cmd, path, options, cb)
     options = nil
   end
   local bufnr = 0
-  lsp.start()
   lsp.client().request("workspace/executeCommand", {
     command = "zk." .. cmd,
     arguments = {
--- a/lua/zk/lsp.lua
+++ b/lua/zk/lsp.lua
@@ -1,8 +1,15 @@
 local config = require("zk.config")
 
-local client_id = nil
-
 local M = {}
+
+local _client
+
+local lsp_get_clients
+if vim.fn.has("nvim-0.10") == 1 then
+  lsp_get_clients = vim.lsp.get_clients
+else
+  lsp_get_clients = vim.lsp.get_active_clients
+end
 
 ---Tries to find a client by name
 function M.external_client()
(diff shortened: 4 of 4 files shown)


tjex:
Great, thanks for this!!
I've since found a bunch of other spots where this occurs, and would go about changing those.
So I'm wondering, why did you choose `vim.tbl_isempty()` instead of `next()`?

It's better to use lua stdlib wherever possible or (for performance, future proof, idiomatics)?

hoofcushion:
It depends on personal preference. `tbl_isempty` has clear semantics and is equivalent to `next`. And using `tbl_isempty` here will not cause performance issue.
I prefer using `next` too, but that's all up to you.

hoofcushion:
`vim.tbl_isempty` API may change when Neovim update, `next` is better.

tjex:
> `vim.tbl_isempty` API may change when Neovim update, `next` is better.

yeah, I was also thinking in this direction. But being unsure whether there was some other magic happening behind the `vim` class.

Then I will go through and change all the other instances of it too and push to your branch. May as well chop them all down while we're at it.

tjex:
ah, that _was_ all of them 😅 Great haha

tjex (commented):
So, as discussed the `next()` implementation is nice.
But I've just tested, and with the `vim.schedule`, marksman (my regular markdown lsp) jumps in as the first lsp when I open a document in my zk notebook.

So it seems like the scheduler is halting our own lsp, to the point where others will be attached in place of it. 

I'm unsure whether this change is then a good change. As it's introduced a bug on my end, where there was none before. And I'm not sure if it will really get in the way if it's left unchanged; as the zk lsp will attach when we open a document within a `ZK_NOTEBOOK_ROOT`, which will fire then once anyway. When it fires, and it it fires multiple times for some reason, I don't think that is a major issue?

hoofcushion:
image
It's necessary for me to use `vim.schedule`. otherwise, it will start zk lsp twice/

tjex:
Ahhh. It looks like you've installed `zk` with mason. 
This is why. zk has its own lsp built in. And looking through the readme both on zk-nvim and zk, this is really not clear enough 😩

So zk lsp from mason is being attached as well as zk's built in lsp...

hoofcushion:
The ideal situation is `zk-nvim` detect the existed Zk LSP client instance and try to attach to it. So `zk-nvim` should wait and see if any client is created.

hoofcushion:
I refactored the code, `vim.schedule` is removed, and I turned off the auto attach option of `zk-nvim` and use `mason` & `lspconfig` to launch `zk lsp` instead.

hoofcushion:
But I think it's better to not set anything manually to use `mason` & `lspconfig` and `zk-nvim` together.

tjex:
I think this is becoming overcomplicated.
There are two options on the user's end:

1. Use zk cli -> do not install zk lsp via Mason / manually
2. Do not use zk cli -> install zk lsp via Mason / manually

I think that instruction is simpler for the user and simpler for code maintenance.

Thanks for the consideration and effort though!

But I'll close for now as, despite the lsp being fired twice in your case (and no doubt for other too), this will not create any adverse behaviour.

> I refactored the code, vim.schedule is removed, as I find out turn off the autoattach option of zk-nvim will avoid the conflict to use mason & lspconfig when launch zk lsp.

Again, the solution is to uninstall Mason's `zk` install.

hoofcushion:
There still has problem with `tbl == {}`, if it tries to get external client, then it has to get it welly.

hoofcushion:
As I said, when use mason, disable `autoattach` for `zk-nvim` will fix the problem 2, since mason & lspconfig will help to set up lsp for `zk-nvim`. It's better for user to not worry about which `zk` to use.

tjex:
I understand. But in the case of the user needing to be aware and make a decision; they still need to include the `autoattach = false` in their config for zk-nvim specifically. So they have to be "extra careful" and make an extra effort anyway.

This means we would change a lot of the code to account for this, and the user would still have to know a finer detail of the setup process and apply that change to their own config. Particularly for people getting to grips with neovim, telling them to implement `opt.lsp.autoattach = false` to their config is more complicated than saying what is now written in the readme (i.e., zk cli = no need for zk lsp installation).

If you want to make a new pr that addresses the `tbl == {}` then I'd be happy to merge that (or go about reverting and cleaning up all the git history in this pr, but that is probably more effort) 🙏

hoofcushion:
done

PR #175, opened 2024-06-26
fix(setup): not really detecting external clients

hoofcushion:


lua/zk/lsp.lua
```diff
@@ -18,16 +18,7 @@ function M.external_client()
     active_clients = vim.lsp.get_active_clients({ name = client_name })
   end
 
-  if active_clients == {} then
-    return nil
-  end
-
-  -- return first lsp server that is actually in use
-  for _, v in ipairs(active_clients) do
```
hoofcushion, line 26: Check if client attached any buffer is not the way to tell if it were in use.

lua/zk/lsp.lua
```diff
@@ -18,16 +18,7 @@ function M.external_client()
     active_clients = vim.lsp.get_active_clients({ name = client_name })
   end
 
-  if active_clients == {} then
```
hoofcushion, line 21: Compare a table to {} is not the way to tell if a table were empty in Lua,

tjex (changes requested):
simply returning `active_clients[1]` is not what we discussed in #173, and it's also not actually the return value that we want. We want to return the `id` of the active client.
It's also error prone anyway as if there were no active clients, it would attempt to return the first element of a table with no elements.

hoofcushion:
> the id of the active client.

Sorry, that is my fault, new commit fixed that.

And active client is not determined by if it attached any buffers. All clients return by the get_clients is active

lua/zk/lsp.lua
tjex: again. this wasn't the agreed upon solution from #173. Simply returning the first active client may introduce a bug, as we want to return _the first client that is actually attached to a buffer_. In #173 you were also implementing a user side solution (your own solution) to disable `auto_attach` in your config. 

If that client was the first in the table, and it was returned then we would be attempting to use a client that wasn't actually attached to the buffer.. Furthermore, there can be active clients attached to different buffers. So returning a client that isn't actually attached to the buffer in question would also be a potential bug enabled by removing this `if` check.

I'm happy for these prs, but for the sake of time, I'm going to push the rest of the changes myself and merge.

Also, I've seen in some of your other PRs that you're using chatGPT. The backwards and forwards we've had with unexpected code changes between comments / prs reminds me a lot of a chatGPT workflow, and has made this all take a lot more time, confused me and made me doubt my decision making. I hope the changes you've been suggesting were not blind suggestions from chatGPT...

tjex (approved):
revert changes: [see comment](https://github.com/zk-org/zk-nvim/pull/175#discussion_r1653791411)

lua/zk/lsp.lua
hoofcushion: I did all the code myself. If my code was bad, I apologize.

> So returning a client that isn't actually attached to the buffer in question would also be a potential bug

I don't understand why a client had no buffer attached could be a potential bug, soon it will attach one as they had well configured, and zk-nvim shouldn't try to create an additional one.

Issue #179, opened 2024-07-18
`ZkNewFromTitleSelection` does not copy last character correctly if it's non-ascii


iAurora:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Same problem #125 reports, in my case the character is Cyrillic. 

The created title ends with � and there is extra code left after the link in original note. 

### How to reproduce?

1. Add a note with content: "Текст заголовка"
2. Visually select "Текст заголовка"
3. Run `:'<,'>ZkNewFromTitleSelection`
4. Check the title of the newly created file
5. Check the link inserted in the original file

<img width="425" alt="Screenshot 2024-07-18 at 05 21 48" src="https://github.com/user-attachments/assets/88d3d22a-dc43-4bc1-9af1-171205804d88">

<img width="425" alt="Screenshot 2024-07-18 at 05 22 17" src="https://github.com/user-attachments/assets/5a03416e-f0c0-465b-b75a-029f110be904">

<img width="425" alt="Screenshot 2024-07-18 at 05 22 37" src="https://github.com/user-attachments/assets/5cd7ec52-47c0-4480-a3b7-d2d7773800cb">

### zk configuration

```toml
# NOTE SETTINGS

# Defines the default options used when generating new notes.
[note]

# Language used when writing notes.
# This is used to generate slugs or with date formats.
language = "ru"

# The default title used for new note, if no `--title` flag is provided.
default-title = "Untitled"

# Template used to generate a note's filename, without extension.
# filename = "{{format-date now '%Y%m%d%H%M%S'}}"
filename = "{{id}}"

# The file extension used for the notes.
extension = "md"

# Template used to generate a note's content.
# If not an absolute path, it is relative to .zk/templates/
template = "default.md"

# Configure random ID generation.

# The charset used for random IDs.
id-charset = "alphanum"

# Length of the generated IDs.
id-length = 4

# Letter case for the random IDs.
id-case = "lower"

# Path globs ignored while indexing existing notes.
ignore = [
  ".obsidian",
	"templates"
]


# EXTRA VARIABLES
#
# A dictionary of variables you can use for any custom values when generating
# new notes. They are accessible in templates with {{extra.<key>}}
[extra]

#key = "value"


# GROUP OVERRIDES
[group.journal]
paths = ["journal"]

[group.journal.note]
filename = "{{format-date now '%Y-%m-%d'}}"
extension = "md"
[...]

PR #187, opened 2024-09-11
Fix get_text_in_range when the last character is multibyte

WhyNotHugo:
The previous implementation would only copy the first byte of the last
character. This is never correct, and produces an invalid sequence.

See: https://github.com/zk-org/zk-nvim/issues/179

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -19,7 +19,7 @@ end)
 
 commands.add("ZkNewFromTitleSelection", function(options)
   local location = util.get_lsp_location_from_selection()
-  local selected_text = util.get_text_in_range(location.range)
+  local selected_text = util.get_selected_text()
   assert(selected_text ~= nil, "No selected text")
 
   options = options or {}
@@ -38,7 +38,7 @@ end, { needs_selection = true })
 
 commands.add("ZkNewFromContentSelection", function(options)
   local location = util.get_lsp_location_from_selection()
-  local selected_text = util.get_text_in_range(location.range)
+  local selected_text = util.get_selected_text()
   assert(selected_text ~= nil, "No selected text")
 
   options = options or {}
@@ -75,7 +75,7 @@ local function insert_link(selected, opts)
   opts = vim.tbl_extend("force", {}, opts or {})
 
   local location = util.get_lsp_location_from_selection()
-  local selected_text = util.get_text_in_range(util.get_selected_range())
+  local selected_text = util.get_selected_text()
 
   if not selected then
     location = util.get_lsp_location_from_caret()
@@ -110,7 +110,7 @@ commands.add("ZkInsertLinkAtSelection", function(opts)
 end, { title = "Insert Zk link", needs_selection = true })
 
 commands.add("ZkMatch", function(options)
-  local selected_text = util.get_text_in_range(util.get_selected_range())
+  local selected_text = util.get_selected_text()
   assert(selected_text ~= nil, "No selected text")
   options = vim.tbl_extend("force", { match = { selected_text } }, options or {})
   zk.edit(options, { title = "Zk Notes matching " .. vim.inspect(selected_text) })
--- a/lua/zk/util.lua
+++ b/lua/zk/util.lua
@@ -100,14 +100,16 @@ function M.get_text_in_range(range)
   local A = range["start"]
   local B = range["end"]
 
-  local lines = vim.api.nvim_buf_get_lines(0, A.line, B.line + 1, true)
-  if vim.tbl_isempty(lines) then
-    return nil
+  local region = vim.region(0, { A.line, A.character }, { B.line, B.character }, vim.fn.visualmode(), true)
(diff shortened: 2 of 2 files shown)


tjex:
Thanks so much for this Hugo, and the extensive detail in the related issue. Really appreciated. I'll need a bit of time to process it all, but I've seen it and am on it!

tjex:
Having a look now. It's awesome to see how much code could be reduced with your solution 🙌 
The byte character issue is (almost) solved, however, there is a new bug being introduced.
Which occurs if white space is included after the last non-whitespace characters in the selection.

This happens both for `:ZkNewFromTitleSelection` and `:ZkNewFromContentSelection`.
If no white space is included after the last non-whitespace chars, then both commands work as expected.

This happens with both ascii and the crylic text:

```text
Pasted from issue.
Текст заголовка 
```

Screenshot 2024-09-18 at 19 09 34

`:ZkNewFromContentSelection {title = "bar"}`

Screenshot 2024-09-18 at 19 11 25

tjex:
Just tried stripping, to no avail (one of my more embarrassing moments, so to say).

tjex:
When returning a plain string, the result is the same:

```lua
  return "Текст заголовка"
```

So, I would say that the way you're selecting the text is fine.

tjex:
Even if there are no white spaces after the text, and the text is selected with visual_line_mode `<V>`, then the error also occurs. Same if the cursor is moved one space further than the end of the text manually, so no white space is actually being selected, just the cursor position is moved. Same deal.

tjex:
Here mentions a way to get the char position instead of the byte position.

Where the `"'<"` etc can be used with any `getpos()` compatible value, which can be char wise or byte wise. I tried quite a few different things. But nothing worked. The residual byte / string chunk is always added to the end of the link if anything further than the text itself is selected in the visual mode.

I've pasted in the below for issue documentations sake, but I'm happy with this solution. 
If we get more gripes about it, or people want to continue off, they can be sent here. But at least the original user who lodged the issue can create _a_ legible link (albeit with a little required cursor precision). Thanks again Hugo!!

`local region = vim.region(0, "'<", "'>", vim.fn.visualmode(), true)`

```help
                                                                *vim.region()*
vim.region({bufnr}, {pos1}, {pos2}, {regtype}, {inclusive})
    Gets a dict of line segment ("chunk") positions for the region from `pos1`
    to `pos2`.

    Input and output positions are byte positions, (0,0)-indexed. "End of
    line" column position (for example, |linewise| visual selection) is
    returned as |v:maxcol| (big number).

    Parameters: ~
      • {bufnr}      (`integer`) Buffer number, or 0 for current buffer
      • {pos1}       (`integer[]|string`) Start of region as a (line, column)
                     tuple or |getpos()|-compatible string
      • {pos2}       (`integer[]|string`) End of region as a (line, column)
                     tuple or |getpos()|-compatible string
      • {regtype}    (`string`) |setreg()|-style selection type
      • {inclusive}  (`boolean`) Controls whether the ending column is
                     inclusive (see also 'selection').

    Return: ~
        (`table`) region Dict of the form `{linenr = {startcol,endcol}}`.
        `endcol` is exclusive, and whole lines are returned as
        `{startcol,endcol} = {0,-1}`.
```


```help
getpos({expr})                                                        *getpos()*
		Get the position for String {expr}.  For possible values of
		{expr} see |line()|.  For getting the cursor position see
		|getcurpos()|.
		The result is a |List| with four numbers:
		    [bufnum, lnum, col, off]
		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
		is the buffer number of the mark.
		"lnum" and "col" are the position in the buffer.  The first
		column is 1.
		The "off" number is zero, unless 'virtualedit' is used.  Then
		it is the offset in screen columns from the start of the
		character.  E.g., a position within a <Tab> or after the last
		character.
		Note that for '< and '> Visual mode matters: when it is "V"
		(visual line mode) the column of '< is zero and the column of
		'> is a large number equal to |v:maxcol|.
		The column number in the returned List is the byte position
		within the line. To get the character position in the line,
		use |getcharpos()|.
		A very large column number equal to |v:maxcol| can be returned,
		in which case it means "after the end of the line".
		If {expr} is invalid, returns a list with all zeros.
		This can be used to save and restore the position of a mark: >vim
			let save_a_mark = getpos("'a")
			" ...
			call setpos("'a", save_a_mark)
<		Also see |getcharpos()|, |getcurpos()| and |setpos()|.
```

```help
getcharpos({expr})                                                *getcharpos()*
		Get the position for String {expr}. Same as |getpos()| but the
		column number in the returned List is a character index
		instead of a byte index.
		If |getpos()| returns a very large column number, equal to
		|v:maxcol|, then getcharpos() will return the character index
		of the last character.

		Example:
		With the cursor on '세' in line 5 with text "여보세요": >vim
			getcharpos('.')		returns [0, 5, 3, 0]
			getpos('.')		returns [0, 5, 7, 0]
```

tjex (approved):

PR #167, opened 2024-05-26
support nvim v0.10.0 lsp api change (get_clients)

tjex:
resolves #165.

lsp nvim <=0.9 -> 0.10: `get_active_clients` -> `get_clients`.

PR #197, opened 2024-10-23
Update to new form of tbl_flatten to fix deprecation warnings

alisonjenkins:
The nightly Neovim is warning that tbl_flatten has been deprecated and should be replaced with what I have put in this commit.

### DIFF
--- a/lua/zk/root_pattern_util.lua
+++ b/lua/zk/root_pattern_util.lua
@@ -8,6 +8,15 @@ local uv = vim.loop
 local M = {}
 
 -- Some path utilities
+
+local function tbl_flatten(...)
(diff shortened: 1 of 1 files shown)


tjex:
Thanks. Have you tested that this works on previous versions of neovim? `vim.iter()` is listed under "new features" for v0.10.

Otherwise, this would need to be implemented and tested with a conditional check like #167

tjex:
Just a heads up that I'm a bit caught up in a job transition AND a transition from mac to linux. I expect to be looking back into zk next week.

tjex (approved):
thanks, good to be here :smiling_imp:. PR all looks good. Thank for refactoring into a function :+1:

PR #210, opened 2025-02-19
feat: add builtin command `ZkNewFromTitleAndContentSelection`

NickvanDyke:
I'm enjoying the plugin and appreciate the builtin commands `ZkNewFromContentSelection` and `ZkNewFromTitleSelection` to refactor and atomize growing notes. However the section I wish to extract frequently already has a markdown heading, e.g.

```markdown
# Cats

...

## The origin of cats

- Nearly all cats, including lions, tigers and the domestic cat, are descended from one prehistoric creature.
- The oldest known relative, the African Wildcat, first developed over 10,000 years ago.
```

So I wrote a new command that will extract the selection's first line as the title (ignoring any starting non-characters like `#` and `-`), and the rest as the content. I think it complements the aforementioned builtins nicely and imagine I'm not the only one with this workflow, hence the PR. Feel free to disagree :)

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -61,7 +61,7 @@ commands.add("ZkNewFromTitleAndContentSelection", function(options)
   assert(selected_text ~= nil, "No selected text")
 
   local title, content = selected_text:match("%W*([^\n]+)\n+(.+)$")
-  assert(title ~= nil and content ~= nil, "No newline-delimited title and content found in selection")
+  assert(title ~= nil or content ~= nil, "No newline-delimited title and content found in selection")
 
   options = options or {}
   options.title = vim.fn.input("Title: ", title)
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -55,6 +55,29 @@ commands.add("ZkNewFromContentSelection", function(options)
   zk.new(options)
 end, { needs_selection = true })
 
+commands.add("ZkNewFromTitleAndContentSelection", function(options)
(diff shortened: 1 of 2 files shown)


lua/zk/commands/builtin.lua
```diff
@@ -55,6 +55,29 @@ commands.add("ZkNewFromContentSelection", function(options)
   zk.new(options)
 end, { needs_selection = true })
 
+commands.add("ZkNewFromTitleAndContentSelection", function(options)
+  local location = util.get_lsp_location_from_selection()
+  local selected_text = util.get_selected_text()
+  assert(selected_text ~= nil, "No selected text")
+
+  local title, content = selected_text:match("%W*([^\n]+)\n+(.+)$")
```
NickvanDyke, line 63: I'm a Lua newbie, there might be a better way to write this pattern. It has worked for me but may miss edge cases that my workflow doesn't hit.

NickvanDyke:
> But I think this feature is too much of an edge case to merit adding it to the code base, which makes the project more complex to maintain and also a larger task to learn for new users.

The implementation has a small scope and the command itself fits nicely with the mental model of the existing builtin refactor commands that users already learn. I can't speak for everyone's zk workflow but this seems to me like a core refactor method.

> This is because you can set the title of a note created with ZkNewFromContentSelection {title = "cutsom title"}. Or if that doesn't suffice, the custom commands API lets users build something like this for themselves.

It's cumbersome to copy/paste or re-type the title. I considered just sharing my custom command in Discussions but thought it might fit the builtin commands.

Anyway, thanks for your work and consideration!

PR #211, opened 2025-02-22
Added the ability to pass down default options to pickers

TewPingz:
**Title:** 
Add support to set default configurations for pickers

**Description:**
Hi there, I've been using the plugin recently, and it's been great! However, I prefer using the Ivy format with Telescope. To allow users more flexibility in customizing their picker, I’ve added support for passing down picker options from the configuration.

This was made in consideration of the other pickers too, users can define default configurations for `fzf_options`, `snacks_picker` and ect.

Here is an example of how this would be used:
```lua
require("zk").setup({
  picker = "telescope",

  picker_options = {
    telescope = {
      layout_strategy = "bottom_pane",
      sorting_strategy = "ascending",
      layout_config = {
        height = 25,
      },
      border = true,
      borderchars = {
        prompt = { "─", " ", " ", " ", "─", "─", " ", " " },
        results = { " " },
        preview = { "─", "│", "─", "│", "╭", "╮", "╯", "╰" },
      },
    },
  },
...
})

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,3 +573,24 @@ require("telescope").load_extension("zk")
 :Telescope zk tags
 :Telescope zk tags created=today
 ```
+
+## Picker Configuration
+
+You can define default configurations for the pickers opened by `zk-nvim`, allowing you to apply a specific theme or layout for `zk-nvim`. This works for all supported pickers, but you'll need to refer to the relevant configuration options for each picker.
+
+```lua
+require("zk").setup({
+    picker_options = {
+        telescope = require("telescope.themes").get_ivy(),
+
+        -- or if you use snacks picker
+
+        snacks_picker = {
+            layout = {
+                preset = "ivy",
+            }
+        },
+    },
+    ...
+})
+```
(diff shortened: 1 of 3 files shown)


tjex:
Nice thought.
So this just overrides the picker options when the picker is called through zk? 

I.e, it doesn't override the picker options globally? Meaning if we change to a non-zk buffer, the picker options will then resort to the default / custom telescope config and not the options set in the zk config?

If so, would you also be able to update the documentation with this change?

TewPingz:
This wont mess with pickers default configuration of the picker, this will only affect the pickers created by `zk-nvim` as they are just passed in as options. This means I can have a theme specific to `zk-nvim` compared to other telescope usages.

Say I define something along the lines of this:
```lua      
require("zk").setup({
        picker = "telescope",
        picker_options = {
          telescope = require("telescope.themes").get_dropdown()
        },
...
}
```

Then I can have that theme for all telescope pickers opened by `zk-nvim` but I will still have the default theme I defined for everything else.

ZK Picker:
image

Telescope find_files
image


For the documentation, I'm thinking of putting it in the Misc section at the bottom, WDYT?

tjex (approved):
Thanks :)

Issue #214, opened 2025-03-01
character_offset error when running ZkNewFromContentSelection (and some other similar commands)


pluma9:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

`:ZkNewFromContentSelection` (and `:ZkNewFromTitleSelection`, `:ZkInsertLinkAtSelection`, `:ZkInsertLink`) produces an error and does not work.

![Image](https://github.com/user-attachments/assets/8a8ea77e-194a-4ae7-a1e7-cb08b33a3959)

> Error executing Lua callback: ...ram Files\Neovim\share\nvim\runtime/lua/vim/lsp/util.lua:2225: attempt to index a nil value
> stack traceback:
> 	...ram Files\Neovim\share\nvim\runtime/lua/vim/lsp/util.lua:2225: in function 'character_offset'
> 	...ram Files\Neovim\share\nvim\runtime/lua/vim/lsp/util.lua:2149: in function 'make_given_range_params'
> 	...\AppData\Local\nvim-data\plugged\zk-nvim/lua/zk/util.lua:47: in function 'get_lsp_location_from_selection'
> 	...al\nvim-data\plugged\zk-nvim/lua/zk/commands/builtin.lua:40: in function 'fn'
> 	...Local\nvim-data\plugged\zk-nvim/lua/zk/commands/init.lua:19: in function <...Local\nvim-data\plugged\zk-nvim/lua/zk/commands/init.lua:12>

### How to reproduce?

Select a piece of text. Run `:ZkNewFromContentSelection`

### zk configuration

```toml
[note]
template = "default.md"

[format.markdown]
hashtags = false
colon-tags = false
multiword-tags = false

[lsp.diagnostics]
dead-link = "error"
```

### Neovim configuration

```viml
require("zk").setup({
	picker = 'telescope'
})
```

### Environment

```bash
zk 0.14.2

NVIM v0.10.3
Build type: Release
LuaJIT 2.1.1713484068


zk-nvim: d113beebac48bf7ad7ce9fdb7f2ce43c90959677


Windows 10
```

tjex:
Just got a somewhat similar error when trying to insert a link:

```text
.local/share/nvim/lazy/zk-nvim/lua/zk/util.lua:104: start_col must be less than or equal to end_col
stack traceback:
        [C]: in function 'nvim_buf_get_text'
[...]

PR #221, opened 2025-04-22
Fix position encoding parameter warning when creating note from selection

vandalt:
When creating a note from title selection with `ZkNewFromTitleSelection`, I keep getting this warning:

```lua
position_encoding param is required in vim.lsp.util.make_given_range_params. Defaulting to position encoding of the first client.
```

It can be fixed by passing the encoding explicitly. For zk, this seems to be utf-16 (`:lua= vim.lsp.get_clients({ name = "zk" })[1].offset_encoding`).

This PR fixes the issue by 1) finding the zk client and 2) setting the encoding to what the zk client returns, or to utf-16 if no zk client is found (but with an error message in the latter case).

I'm not sure if the whole encoding detection part is necessary. Could we just hardcode "utf-16" as an argument to `make_given_range_params` directly or could this change depending on ZK installations?

Thanks!

### DIFF
--- a/lua/zk/util.lua
+++ b/lua/zk/util.lua
@@ -53,6 +53,7 @@ local function get_offset_encoding(bufnr)
       string.format("ZK Client (id: %s) offset_encoding is nil. Do not unset offset_encoding.", zk_client.id),
       error_level
     )
+  else
     offset_encoding = zk_client.offset_encoding
   end
   return offset_encoding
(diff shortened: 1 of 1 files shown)


tjex:
Thanks. Potentially helps with #214 .

Note: I'm very time poor until July and can't say exactly when I can look at merging this or anything else for that matter. Hopefully one of the other maintainers checks back in in that time.

PR #224, opened 2025-05-02
vim.validate update

mainendra:
Getting a deprecation waring for vim.validate function. Used type to avoid that.


mainendra:
@tjex - Yes, ideally we need to use `vim.validate('func', func, 'function')` but it's for neovim 0.11 and above. do you want to add neovim version check or update supported neovim version?

tjex:
Sorry, forgot to tag / comment here. I added it in #226 .
We're now only supporting neovim 0.11.* for future updates of zk-nvim

PR #227, opened 2025-05-09
Fix position encoding warning from deprecated method

tjex:
This PR all @vandalt from #221 !!

I merged his pr too quick and had to revert (#225) after finding that it broke neovim 0.10.4.

So am getting his code back in myself (as the current HEAD is now only supporting neovim v0.11.0 and up) instead of pestering him to make up for my merge trigger finger...


vandalt:
Sorry about the trouble, I forgot to test on older Neovim versions. Thanks for handling this @tjex!

Issue #228, opened 2025-05-09
"client_id: expected number, got nil" when opening a note


heyjustin:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Clean install of neovim with zk plugin (managed using lazy)

**Neovim**
```
NVIM v0.11.1
Build type: Release
LuaJIT 2.1.1744318430
Run "nvim -V1 -v" for more info
```

**zk-nvim**
```
return {
  "zk-org/zk-nvim",
  config = function()
    require("zk").setup({})
  end
}
```

**Local zk-vim git rep**
```
zk-nvim git:(main) git log --oneline |head
69e6f9c Fix position encoding warning from deprecated method (#227)
e9cb2b3 vim.validate update (#224)
9597048 update deprecated api methods and remove nvim-0.10 version check (#226)
```

When opening a note, I get the following error:
```
Failed to run `config` for zk-nvim

.../Cellar/neovim/0.11.1/share/nvim/runtime/lua/vim/lsp.lua:941: client_id: expected number, got nil

# stacktrace:
  - vim/shared.lua:0 _in_ **validate**
  - /opt/homebrew/Cellar/neovim/0.11.1/share/nvim/runtime/lua/vim/lsp.lua:941 _in_ **buf_attach_client**
  - /zk-nvim/lua/zk/lsp.lua:45 _in_ **buf_add**
  - /zk-nvim/lua/zk.lua:34 _in_ **_lsp_buf_auto_add**
  - /zk-nvim/lua/zk.lua:18 _in_ **setup_lsp_auto_attach**
  - /zk-nvim/lua/zk.lua:44 _in_ **setup**
  - ~/.config/nvim/lua/plugins/zk.lua:4 _in_ **config**
  - ~/.config/nvim/lua/config/lazy.lua:19
  - ~/.config/nvim/init.lua:4
```

Willing to help with the problem. Very new to lua

### How to reproduce?

1. `brew install neovim`
2. `brew install zk`
3. Configure lazy
4. Configure  zk-nvim
5. Create a notebook (`zk init`)
6. `zk new`

### zk configuration

```toml
[notebook]
dir = "$HOME/Documents/notes/main"

[note]
language = "en"
filename = "{{format-date now \"timestamp\"}}-{{slug title}}"
extension = "md"
template = "default.md"
```

### Neovim configuration

```viml
**init.lua**

[...]

PR #229, opened 2025-05-10
add some guards to lsp client_id use

JulianNymark:
when i ran `zk new` from the cli, I would get this error in nvim, it seems to connect to the lsp and work eventually, but maybe this is an ordering thing? (I'm out of my element here, but maybe the lsp takes some time to boot up or such?). 

It's probably nice to have these guards in place regardless? They show up as warnings with lua diagnostics 🤔.

### DIFF
--- a/lua/zk/lsp.lua
+++ b/lua/zk/lsp.lua
@@ -42,7 +42,9 @@ end
 function M.buf_add(bufnr)
   bufnr = bufnr or 0
   M.start()
-  vim.lsp.buf_attach_client(bufnr, client_id)
+  if client_id then
+    vim.lsp.buf_attach_client(bufnr, client_id)
+  end
 end
 
 ---Stops the LSP client managed by this plugin
@@ -56,7 +58,9 @@ end
 
 ---Gets the LSP client managed by this plugin, might be nil
 function M.client()
-  return vim.lsp.get_client_by_id(client_id)
+  if client_id then
+    return vim.lsp.get_client_by_id(client_id)
+  end
 end
 
 return M


tjex:
What is the error you get? I think you forgot to include it?

JulianNymark:
> What is the error you get? I think you forgot to include it?

ah, my bad, here's the exact error:

```
Failed to run `config` for zk-nvim

.../Cellar/neovim/0.11.0/share/nvim/runtime/lua/vim/lsp.lua:912: client_id: expected number, got nil

# stacktrace:
  - vim/shared.lua:0 _in_ **validate**
  - /opt/homebrew/Cellar/neovim/0.11.0/share/nvim/runtime/lua/vim/lsp.lua:912 _in_ **buf_attach_client**
  - /zk-nvim/lua/zk/lsp.lua:45 _in_ **buf_add**
  - /zk-nvim/lua/zk.lua:34 _in_ **_lsp_buf_auto_add**
  - /zk-nvim/lua/zk.lua:18 _in_ **setup_lsp_auto_attach**
  - /zk-nvim/lua/zk.lua:44 _in_ **setup**
  - ~/.config/nvim/lua/plugins/notetaking.lua:5 _in_ **config**
  - ~/Repos/configs/nvim/init.lua:96
```

tjex:
Ah ok. This is same as #228 

How have you installed zk lsp? Through neovim, the cli or both?

heyjustin (approved):
Applied these changes locally in response to #228  - and can confirm that it resolves the issue.

tjex:
The code changes will hide the error, but they will also remove any hints at what is going wrong. It's not necessarily a great fix. Either an `else` statement should be added stating why that branch has been reached, or it should be handled in some other way.

I've just tested on another machine with nvim 0.11.1, HEAD of zk-nvim, and zk 0.15.1 and I'm not getting this error.

tjex:
Are you configuring zk anywhere outside of zk's own lazy config? e.g, lsp config?

JulianNymark:
Yeah, i have the following zk config (that gets symlinked to `~/.config/zk`): https://github.com/JulianNymark/configs/blob/master/zk/config.toml

Neovim: 0.11.1
zk-nvim: commit 69e6f9c
zk: 0.15.1
on Macos

tjex:
Hmm, well this sucks, because I'm running (almost) the exact same combo. I'm just on linux instead. But that shouldn't be the issue here... Ah man, I get that this pr will 'fix' the issue, but it could easily make issues further down the track much more abstract to fix. 

Will need some more time to investigate.

JulianNymark:
I've tried to use this sample-init.lua now: https://github.com/JulianNymark/zk-nvim/blob/sample-init/sample-init.lua

Should be fairly minimal, all it has is lazy + zk-nvim.

updated my zk config to use it by setting:

```
editor = "nvim -u ~/Repos/zk-nvim/sample-init.lua"
```

Seems to still have this bug 😵‍💫

tjex:
You're right. Ok, then, will keep having a look at this when I can.

Can you paste here the output of `:LspInfo` after you've opened a zk note _with_ your fix applied. I'm interested to see what is actually getting attached. I suspect zk isn't being attached at all. Because that's what this is doing:

```lua
  if client_id then
    vim.lsp.buf_attach_client(bufnr, client_id)
  end
```

As there is no client, nothing will be attached.
So no error, but also no zk.

tjex:
I'm pretty sure it's stemming from the change from `vim.tbl_flatten()` (deprecated function of neovim 0.10) to `vim.iter({ ... }):flatten():totable()`, in #197 and #226 (where I removed the checks for neovim version.

When I run the minimal config with nvim version 0.10.4, there is also no client, but the error message complains about notebook_pattern_util getting a string instead of a table. 

```text
Failed to run `config` for zk-nvim                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                    
...nvim/data/nvim/lazy/zk-nvim/lua/zk/root_pattern_util.lua:82: opt: expected table, got string                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                    
# stacktrace:                                                                                                                                                                                                                                                                                                                                       
  - vim/shared.lua:0 _in_ **validate**                                                                                                                                                                                                                                                                                                              
  - /zk-nvim/lua/zk/root_pattern_util.lua:82 _in_ **notebook_root**                                                                                                                                                                                                                                                                                 
  - /zk-nvim/lua/zk.lua:30 _in_ **_lsp_buf_auto_add**                                                                                                                                                                                                                                                                                               
  - /zk-nvim/lua/zk.lua:18 _in_ **setup_lsp_auto_attach**                                                                                                                                                                                                                                                                                           
  - /zk-nvim/lua/zk.lua:44 _in_ **setup**                                                                                                                                                                                                                                                                                                           
  - ~/scratch/init-zk/init.lua:25 _in_ **config**                                                                                                                                                                                                                                                                                                   
  - ~/scratch/init-zk/init.lua:21         
```

JulianNymark:
I'm guessing the PR https://github.com/zk-org/zk-nvim/pull/230, fixes this in a better way than this one right?

> Can you paste here the output of :LspInfo

```
...
vim.lsp: Active Clients ~
- zk (id: 1)
  - Version: 0.15.1
  - Root directory: nil
  - Command: { "zk", "lsp" }
  - Settings: {}
  - Attached buffers: 1
...
```

Strange that it works 🤔 (lsp is up and running). I tested  https://github.com/zk-org/zk-nvim/pull/230, and it doesn't have the message, but also `:LspInfo` prints the following:

```
...
vim.lsp: Active Clients ~
- No active clients
...
```

I might have rushed this a bit... hastily done during lunch break 😅 (i'll check more later)

tjex:
Huh, wow. It's true.. Great catch. Turns out we now need to call `vim.lsp.enable('zk')`.

Not sure yet whether to add this in zk lsp itself, or in the neovim config. Will do some investigating. But in any case, yes, #230 is still the pr that should be merged as a fix for this.

But thanks all the same. You helped a lot!

Issue #68, opened 2022-08-03
How can i search content using telescope ui?


zsmatrix62:
Hi,

Im quite new to lua development, so wondering how can i do a live-searching in telescope ui? jus tlike the one in searching ZKNotes?

kylo252:
Here's a simple implementation for `ZkGrep`

```lua
function M.grep_notes(opts)
  local collection = {}
  local list_opts = { select = { "title", "path", "absPath" } }
  require("zk.api").list(vim.env.ZK_NOTEBOOK_DIR, list_opts, function(_, notes)
    for _, note in ipairs(notes) do
      collection[note.absPath] = note.title or note.path
    end
  end)
  local options = vim.tbl_deep_extend("force", {
    prompt_title = "Notes",
    search_dirs = { vim.env.ZK_NOTEBOOK_DIR },
    disable_coordinates = true,
    path_display = function(_, path)
      return collection[path]
    end,
  }, opts or {})
  require("telescope.builtin").live_grep(options)
end

vim.api.nvim_create_user_command("ZkGrep", M.grep_notes, {})

```

Note: it's possible to make the display closer to `zk edit`, but that requires re-writing a complex `entry_maker.gen_from_vimgrep`

zsmatrix62:
Thanks! this totally works!!

riodelphino:
Hi, I've enhanced @kylo252 's code.

Modified:
* Change zk dir automatically by `get_root()` func
* Add `entry_maker()` to modify list displaying text
* Add `displayer()` to show the list highlighted
* Add `ordinal` as title, lnum, col, text (working??)

Screenshot:
<img width="1042" alt="Image" src="https://github.com/user-attachments/assets/6d4d8146-64f8-4767-90af-1eb73e74e07f" />


ZkGrep with highlighted list (telescope):
```lua
local util = require('zk.util')

local function get_root()
   return util.notebook_root(vim.fn.expand('%:p:h'))
end

function grep_notes(opts)
   local collection = {}
   local root = get_root()
   local list_opts = { select = { 'title', 'path', 'absPath' } }
   require('zk.api').list(root, list_opts, function(_, notes)
      for i, note in ipairs(notes) do
         collection[note.absPath] = note.title or note.path
      end
   end)
   local displayer = require('telescope.pickers.entry_display').create({
      separator = ' ',
      items = {
         -- if no limit title chars
         {}, -- title col
         {}, -- text col
         -- if limit title chars
[...]

PR #232, opened 2025-05-15
Zk grep 2 (inheriting #171)

riodelphino:
Initially, I added [this sample code](https://github.com/zk-org/zk-nvim/issues/68#issuecomment-2883787761) to implement the `:ZkGrep` command.

Then I merged the changes from #171 into the latest `main` branch, and extended it based on the above example.

### Features:
- Grep with telescope and snacks.picker
- Highlighted entries for better readability
- Jumps directly to the matched line
- Above features are solved by using `rg` command, instead of `zk list --format=oneline -m` command. (to get `lnum` and `col`)

### TODO:

- [ ] Sorting order is incorrect in snacks_picker
Now `score -> idx`.
Sorting by `score -> title(filename) -> lnum -> col` or modifying scoring directly is ideal.
 snacks_picker's sorting is difficult to understand, so I give up for now. 
[This may helps (Japanese)](https://minerva.mamansoft.net/Notes/%F0%9F%93%9C2025-04-12+obsidian.nvim%E3%81%A7snacks.picker%E3%82%92%E4%BD%BF%E3%81%A3%E3%81%9F%E3%81%A8%E3%81%8D%E3%81%ABObsidianQuickSwitch%E3%81%A7%E8%A1%A8%E7%A4%BA%E3%81%95%E3%82%8C%E3%82%8B%E5%80%99%E8%A3%9C%E3%82%92%E6%9C%80%E8%BF%91%E9%96%8B%E3%81%84%E3%81%9F%E3%83%95%E3%82%A1%E3%82%A4%E3%83%AB%E9%A0%86%E3%81%AB%E3%81%99%E3%82%8B)

Thanks to @shfattig for providing a great foundation and roadmap for integrating grep into `zk-nvim` pickers.


riodelphino:
Optional highlight groups for ZkGrep entries.
Please add this to your config to customize the colors.
(These colors are from `zk list -i` command on my terminal.)

init.lua:
```lua
vim.api.nvim_create_autocmd('ColorScheme', {
  callback = function()
    vim.api.nvim_set_hl(0, 'ZkGrepEntryTitle', { fg = '#E0B449', bold = true })
    vim.api.nvim_set_hl(0, 'ZkGrepEntryText', { fg = '#E4E4E4' })
  end,
})
```

riodelphino:
ChangeLog:
- Use built-in hl groups 'TelescopeResults*' instead of custom 'ZkGrep*'
- Updated README.md to reflect these changes

Now using:
- `TelescopeResultsIdentifier` for title
- `TelescopeResultsNormal` for body text

To customize these highlights, you can add something like this to your config:
```lua
vim.api.nvim_create_autocmd('ColorScheme', {
  callback = function()
    vim.api.nvim_set_hl(0, 'TelescopeResultsIdentifier', { fg = '#E0B449', bold = true })
    vim.api.nvim_set_hl(0, 'TelescopeResultsNormal', { fg = '#E4E4E4' })
  end,
})
```

riodelphino:
ChangeLog:
- Use built-in hl groups 'TelescopeResults*' instead of custom 'ZkGrep*'
- Updated README.md to reflect these changes

Now using:
- `TelescopeResultsIdentifier` for title
- `TelescopeResultsLineNr` for lnum:col
- `TelescopeResultsNormal` for body text

To customize these highlights, you can add something like this to your config:
```lua
vim.api.nvim_create_autocmd('ColorScheme', {
  callback = function()
    vim.api.nvim_set_hl(0, 'TelescopeResultsIdentifier', { fg = '#E0B449', bold = true })
    vim.api.nvim_set_hl(0, 'TelescopeResultsLineNr', { fg = '#999999' })
    vim.api.nvim_set_hl(0, 'TelescopeResultsNormal', { fg = '#CCCCCC' })
  end,
})
```

tjex:
This is awesome, thanks for the work. I really want to get grep functionality in here. I'm sad that #171 stalled out. Using zk's `match` was great because it's insanely fast (faster than `rg`), but without jumping to or highlighting the results, it came at a price.

All that being said, I have to prioritise bug fixes much more highly at the moment as until the start of July I need to finish my masters thesis, while working part time on the side. So just focusing on bug fixes of existing functionality keeps my other commitments manageable, while keeping the ship afloat here.

But I am totally in to get this merged, and will look at it when I can, but latest in July.
Unless one of the other maintainers pokes their head back in. But it's getting quieter and quieter.

riodelphino:
Thank you for replying even though you're so busy.
And I'm glad that you see value in this.

Yes, `zk list` is probably quite fast.
If only that command could return lnum and col, it would be ideal.

Of course, please prioritize bug fixes — I’d be happy if you could take a look when you have some spare time.

In the meantime, I might continue adding commits here with further improvements.
See you in July or later!


### Concerns (for future reference):
Currently, ZkGrep is integrated into `ui.pick_notes()`.
However, this makes `ui.pick_notes()` more complex due to the added `if-then-else` logic.
I'm considering whether it would be better to create a separate function, such as `ui.pick_grep_notes()`, instead.

riodelphino:
And now I'm looking into integrating `grep` with `snacks.picker`.
I had a hard time understanding snacks.nvim's code.
I'll probably add a commit soon.

riodelphino:
It works!

- Add `grep` to `snacks_picker` not only `telescope`
- Merged `bufname_from_yaml` branch


But I'm a bit confused by git's merged commit history — the order looks upside down to me...

### An Issue:
In `snacks_picker`, lines containing `:` — such as
`this is the error line 24:00:00`
— cause a parsing error in the `rg` command.

The file path is incorrectly parsed as
`path/to/file.md:100:1:this is the error line 24`

telescope's grep does not have same issue. Because the parsing regex was modifiable from outside.

`snacks_picker` need to be forked to recieve `--json` option to `rg` command, or `parse_regex` option.  I think these solve this issue.

riodelphino:
Thanks a lot for your guidance !!

My TODO:
- Add to `zk/doc.txt`.
- Fix the slow display on bufferline with too many files.
- Add Grep feature to fzf's live_grep ? (or not ?)

It will take some time.

riodelphino:
## Enabled grepping another notebook

I think above [issue ](https://github.com/zk-org/zk-nvim/pull/232#issuecomment-3168385709) was solved.
Now `:ZkGrep` and `zk.grep_notes()` also can show another zk notebook.

```vim
:ZkGrep { notebook_path = 'path/to/another_zk_notebook'}
```
```lua
require('zk').grep_notes( { notebook_path = 'path/to/another_zk_notebook'} )
```


### Splitted grep and pick

To avoid complexity, `grep` and `pick` are separated.
`zk.edit` also checks if the `options` include `{ grep = true }` or not.


### Unacceptable changes?

I made many modifying in this commit with trial and error.
Please let me know if there are unacceptable changes.

riodelphino:
Still need to modify `zk.txt` and `README.md`.

riodelphino:
Oops, I accidentally included the YAML sample codes and documents that should have been part of the #237 PR (in README.md and zk.txt).

(But it should be fine even if they are included here.)

tjex:
But neotree-zk hasn't been active for a long time. Probably better to support the upstream?
Makes it complicated for users to have to install a special flavour of neotree to use zk?

riodelphino:
> But neotree-zk hasn't been active for a long time. Probably better to support the upstream?

Yes, neotree-zk should support upstream neo-tree, because it currently throws some errors. (But basically it works.)

> Makes it complicated for users to have to install a special flavour of neotree to use zk?

No, I don't think so, because my sample configs for neo-tree (and bufferline) are quite long.

I think it’s easier for users to install a dedicated flavor like neotree-zk and just add some keymaps.

riodelphino:
Anyway, now I think my sample codes for neo-tree and bufferline are outside of the scope of this PR.
To avoid delaying the merge, I'm going to remove it for now.

(I merged the upstream changes to this branch. Is it a correct behavior?)

tjex:
Hey. Thanks for all the work here. I'm seeing it all, just in the last week of madness. Have my thesis presentation next monday, and then on holiday for a few weeks. So will be looking at this and the custom buffer titles PR from next week on :D

riodelphino:
Humm... sooooo dificult...

### Async or Sync
`zk.api.list` is async, but `bufferline.nvim` does not accept async.
It costs toom many codings to bridge between them.

### Not accepts single note path
`zk list` cli command accepts notebook dir and also single file path (This is OK)
But `zk.api.list` or zk lsp accept only notebook dir and returns all file's infomation. So need to search all files every time. (probably it's an issue)

### Returns broken JSON
`zk list --format json {file_path} | jq ...` always returns JSON with `Found {n} note` text in the top or the bottom. It breaks JSON format.
The [zk document](https://zk-org.github.io/zk/) does not seem to discribe how to disable the unnecessary text message.

Using `lyaml` is so much easier.

riodelphino:
This is the code for it.
Which do you think is better, `lyaml` version #237 or this api version ???

### Concepts
- Using builtin `zk.list` api (No need to implement `lyaml`)
- vim.b.zk_title keeps the title
- vim.o.zk_list table keeps the `zk list` output as JSON (`zk.list` api can't load a single note. So have to keep the whole list at once.)
- Update zk_list and zk_title in `BufWritePost`, then refresh bufferline

### Code
bufferline config with lazy.nvim:
```lua
config = function()
   require('bufferline').setup({
      -- ...
      -- other configs
      -- ...
      name_formatter = function(buf)
         local ok, zk_title = pcall(vim.api.nvim_buf_get_var, buf.bufnr, 'zk_title')
         if ok and zk_title then
            return zk_title
         else
            vim.schedule(function()
               update_zk_list(function()
                  if refresh_title(buf) then require('bufferline.ui').refresh() end
               end)
            end)
         end
         return vim.fn.fnamemodify(buf.name, ':t:r')
      end,
   })

   -- Refresh Buffer Title
   function refresh_title(buf, callback)
      -- Parse title
      if vim.g.zk_list then
         for _, note in ipairs(vim.g.zk_list) do
            if note.absPath == buf.path then
               local title = note.title or note.filenameStem or note.id
               vim.api.nvim_buf_set_var(buf.bufnr, 'zk_title', title)
               if callback then vim.schedule(callback) end
               return true
            end
         end
      end
      return false
   end
   
   -- Async update zk_list
   function update_zk_list(callback)
      require('zk.api').list(nil, {
         select = { 'id', 'absPath', 'title', 'filenameStem' },
      }, function(err, notes)
         if not err and notes then
            vim.g.zk_list = notes
            print('zk_list updated')
         else
            print('zk_list update error:', err)
         end
   
         if callback then vim.schedule(callback) end
      end)
   end
   
   -- Add BufWritePost autocmd
   vim.api.nvim_create_autocmd('BufWritePost', {
      pattern = '*.md',
      callback = function()
         print('autocmd triggered')
         local bufnr = vim.api.nvim_get_current_buf()
         local path = vim.api.nvim_buf_get_name(bufnr)
         local buf = { bufnr = bufnr, path = path }
   
         -- Clear existing zk_title
         vim.b[bufnr].zk_title = nil
   
         -- Update zk_list -> update zk_title -> refresh
         update_zk_list(function()
            refresh_title(buf, function()
               require('bufferline.ui').refresh()
               print('autocmd completed: bufferline refreshed')
            end)
         end)
      end,
   })
end,
```

riodelphino:
After some investigations, now I understand that we don't need `lyaml` here.
`zk.list` is enough for that purpose.

So I will replace the YAML fetching in this grep feature, with `zk.list`.

riodelphino:
> The count is because `less` is being used as a pager to display the results. You need to pass the `--quiet` option.

Got it -- there's an option to hide the count!

riodelphino:
Thanks for reviewing !

Additionally, I created PR in `neo-tree-zk`, for syncing upstream.
https://github.com/zk-org/neo-tree-zk.nvim/pull/2

Though it took quite a while, it's pretty close to completion, too.

But get some rest first! ☺️

riodelphino:
I’ve merged the latest upstream main into this branch due to conflicts and avoiding deviation.
Is this the correct approach for a GitHub PR?
(I hope this doesn't cause any extra work for the reviewers..)

tjex:
Yep, thats the correct process. And main is also correct (on zk, we have a dev branch which is the default for merges, but zk-nvim is still just main). I know this is taking ages to get this merged. Sorry for it and thanks for your patience. 

@WhyNotHugo Do you think you'd have any time coming up to reassess now that riodelphino replied to your questions? I agree it would be great to use zk's match, but without line / location jumping, its not great. So it is good that we could switch out ripgrep for zk match (I haven't varified how much of a rewrite that would take though).

tj-kolleh-ck:
Is this feature still needed/wanted?

riodelphino:
@tj-kolleh-ck
I'm actively using this feature in my fork on a daily basis.
So it has been quite useful in practice, for me.

tjex:
It's wanted in general, but this implementation couples the snacks picker too tightly. We need to interface with `vim.ui.select()` in reality. Supporting all the pickers is just not sustainable. It's also true that building in ripgrep as a dependency is just another thing to maintain, when really we should just work on returning locations from zk match.

riodelphino:
@tjex
Thanks.
Your conversation with @WhyNotHugo seemed to be dropped off, but now your reply made me realize that you both are sharing the same opinion about this PR.

I understand that the followings are deprecated in zk-nvim:
- Depending on ripgrep or other external tools (even temporarily).
- Implementing features using other plugins.

So, I will close this PR👍️

Although I'm not sure how to implement an interactive grep feature using only `vim.ui.select`, I'm looking forward to `zk match` returns locations☺️

### My small suggestion:
It would be helpful to add a warning like the above to the `README.md`.

Otherwise, some zk-nvim users (like me) might try to create PRs like this and spend time and effort on them, since we can see that zk-nvim **already** integrates other picker plugins.

PR #220, opened 2025-04-20
feat: create new note from fzf_lua notes picker

mhanberg:
## Description

This makes it possible to create a new note from the note pickers. Currently only implemented for fzf_lua because that is what I use.

CleanShot 2025-04-20 at 19 05 54@2x

If this patch doesn't get merged, here is an example to get this behavior from your own config.

```lua
  local zk = require("zk")
  local ansi_codes = require("fzf-lua").utils.ansi_codes

  local opts = function(tbl)
    return vim.tbl_extend("keep", { buffer = 0, silent = true }, tbl)
  end

  vim.keymap.set("n", "<space>zf", function()
    zk.edit({ sort = { "modified" } }, {
      fzf_lua = {
        fzf_opts = {
          ["--header"] = ansi_codes.blue("CTRL-E: create a note with the query as title"),
        },
        actions = {
          ["ctrl-e"] = function()
            local query = require("fzf-lua").config.__resume_data.last_query
            require("zk").new { title = query }
          end,
        },
      },
    })
  end, opts { desc = "Find notes" })
```

Related to #200

PR #245, opened 2025-08-14
feat: create new note from telescope picker

benjamineeckh:
This achieves the same functionality as #220, but now with telescope.
image

As mentioned in #220 this allows for the same functionality from a config:
```lua
local zk = require("zk")
local actions = require("telescope.actions")
local action_state = require("telescope.actions.state")

local opts = function(tbl)
  return vim.tbl_extend("keep", { buffer = 0, silent = true }, tbl)
end

vim.keymap.set("n", "<space>zf", function()
  zk.edit({ sort = { "modified" } }, {
    telescope = {
      prompt_title = "CTRL-E: create a note with the query as title",
      results_title = "Create new file or edit existing",
      attach_mappings = function(prompt_bufnr, mapping)
        mapping("i", "<C-e>", function()
          local current_picker = action_state.get_current_picker(prompt_bufnr)
          local prompt = current_picker:_get_prompt()
          actions.close(prompt_bufnr)
          vim.schedule(function()
            require("zk").new({ title = prompt })
          end)
        end)
        mapping("i", "<CR>", function()
          local entry = action_state.get_selected_entry()
          if entry == nil then
            actions.close(prompt_bufnr)
          else
            vim.schedule(function()
              actions.select_default(prompt_bufnr)
            end)
          end
        end)
        return true
      end,
    },
  })
end, opts({ desc = "Find notes" }))
```

### DIFF
--- a/lua/zk/pickers/telescope.lua
+++ b/lua/zk/pickers/telescope.lua
@@ -59,7 +59,11 @@ end
 
 function M.show_note_picker(notes, options, cb)
   options = options or {}
-  local telescope_options = vim.tbl_extend("force", { prompt_title = options.title }, options.telescope or {})
+  local telescope_options = vim.tbl_extend(
+    "force",
+    { prompt_title = options.title },
+    options.telescope or { prompt_title = "CTRL-E: create a note with the query as title" }
+  )
 
   pickers
     .new(telescope_options, {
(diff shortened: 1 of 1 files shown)


tjex:
sorry for the delay, and thanks for the patience! tested and working.

PR #261, opened 2025-09-18
Pass ZkTags options on to zk.edit

rcmosher:
Pass `options` from `:ZkTags` on to `zk.edit()`, preserving `tags` selected by the user. This allows `zk.edit()` to maintain the `notebook_path` used in looking up tags.

Please let me know if there are options that should be filtered out.

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -127,6 +127,7 @@ commands.add("ZkTags", function(options)
     tags = vim.tbl_map(function(v)
       return v.name
     end, tags)
-    zk.edit({ tags = tags }, { title = "Zk Notes for tag(s) " .. vim.inspect(tags) })
+    options = vim.tbl_extend("keep", { tags = tags }, options or {})
+    zk.edit(options, { title = "Zk Notes for tag(s) " .. vim.inspect(tags) })
   end)
 end)


tjex:
This is a good find thanks. But this behaviour is then present elsewhere. E.g, `:ZkNotes {notebook_path = "/path"}` and then creating a new note with `CTRL-E`. 

That note will be created in the global notebook path.

It would be great, if we want to support opening/creating notes from secondary notebooks than the global notebook, that we address all cases, and not a singular use case.

Would you be up for addressing that as well?

rcmosher:
I'll take a look. I see I need to at least update telescope and fzf_lua.

lua/zk/pickers/fzf_lua.lua
```diff
@@ -69,7 +69,11 @@ function M.show_note_picker(notes, options, cb)
       end,
       ["ctrl-e"] = function()
         local query = require("fzf-lua").config.__resume_data.last_query
-        require("zk").new({ title = query })
+        local newOptions = { title = query }
+        if options["notebook_path"] then
+          newOptions["notebook_path"] = options["notebook_path"]
+        end
+        require("zk").new(newOptions)
```
tjex: lua is written in snake case: `new_options`

lua/zk/pickers/telescope.lua
```diff
@@ -93,9 +93,13 @@ function M.show_note_picker(notes, options, cb)
...
           local current_picker = action_state.get_current_picker(prompt_bufnr)
           local prompt = current_picker:_get_prompt()
+          local newOptions = { title = prompt }
+          if options["notebook_path"] then
+            newOptions["notebook_path"] = options["notebook_path"]
+          end
           actions.close(prompt_bufnr)
           vim.schedule(function()
-            require("zk").new({ title = prompt })
+            require("zk").new(newOptions)
```
tjex: same comments as in fzf_lua

tjex (changes requested):

lua/zk/pickers/fzf_lua.lua
```diff
@@ -69,7 +69,11 @@ function M.show_note_picker(notes, options, cb)
       end,
       ["ctrl-e"] = function()
         local query = require("fzf-lua").config.__resume_data.last_query
-        require("zk").new({ title = query })
+        local newOptions = { title = query }
+        if options["notebook_path"] then
+          newOptions["notebook_path"] = options["notebook_path"]
+        end
```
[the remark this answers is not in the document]
rcmosher: Just opted to make your suggested change to pass on the existing options plus title.

tjex (approved):
Thanks :)

PR #265, opened 2025-11-01
fzf_lua: fix multi_select, requires passing --multi

nmeum:
Without passing `--multi`, zk's `multi_select` option does not work in conjunction with the `fzf_lua` picker.

### DIFF
--- a/lua/zk/pickers/fzf_lua.lua
+++ b/lua/zk/pickers/fzf_lua.lua
@@ -96,6 +96,7 @@ function M.show_tag_picker(tags, options, cb)
       ["--nth"] = 2,
       ["--exact"] = "",
       ["--tabstop"] = 4,
+      ["--multi"] = options.multi_select,
     },
     actions = {
       ["default"] = function(selected, _)


tjex:
What are you meaning by the `multi_select` option? A reproduction example would be useful, as my initial idea is to open fzf_lua picker in neovim, select multiple notes, and open. But this does open multiple notes.

nmeum:
Hi! `multi_select` is an option that can be passed to the picker abstraction provided by zk-nvim, see:

https://github.com/zk-org/zk-nvim/blob/8fdd7032633045ece559273370fc0ec75ee8ffce/doc/zk.txt#L378-L383

The `multi_select` option can be used to select multiple items at once, depending on the utilized picker. An example usage of the `pick_tags` function from above would be:

```
:lua require("zk").pick_tags({}, {}, vim.print)
```

With this patch applied, and the `fzf_lua` picker, you can select multiple items using `Shift+Tab` without this patch, this doesn't work because zk-nvim doesn't pass `--multi` to `fzf_lua`. This is an existing functionality of zk-nvim, it is just not correctly implemented for the `fzf_lua` picker.

tjex:
Ok, but _without_ this patch I _can_ select multiple items in the `fzf_lua` picker...

`:ZkNotes -> select multiple -> enter -> multiple notes are opened` 

Picking multiple tags isn't supported, in case that's what you're trying. 
`:lua require("zk").pick_tags({}, {}, vim.print)` with your patch, and the current HEAD print the exact same thing. In other words, your patch isn't changing any behaviour for me.

Sorry if I'm missing something here, but I'm a bit stumped as to where I could be getting this wrong.

PR #275, opened 2025-11-29
feat: Cache zk info to buffer variable

riodelphino:
Related to [#260](#260) 

Sorry for many PRs.
But I've heard that splitting code and PR helps merging.

### Purpose

Make it easier to implement features that need per-buffer zk info.

(e.g.)
-  integrate bufferline [#260](#260) 
- which-key buffers list

### Details

This PR is a ground work.

It adds a feature to cache `zk.api.list()` results in a buffer-local variable.
They can be accessed easily by `vim.b.zk` or `vim.b[bufnr].zk`.

The cache is fetched & updated in these events:
VimEnter — fetches the cache for all buffers on startup
BufReadPost — fetches the cache on read
BufWritePost — updates the cache on save

#### Start nvim (VimEnter):
1. Open `nvim` with buffers: `nvim file1.md file2.md`
2. Or open buffers automatically by a session manager plugin
3. Then `:lua for _, buf in ipairs(vim.api.nvim_list_bufs()) do name = vim.api.nvim_buf_get_name(buf) if name:match('%.md$') then print(vim.inspect(vim.b[buf].zk)) end end`
4. Confirm all buffers already have zk info

#### Open a buffer (BufReadPost):
1. Open a buffer
2. Then `:lua print(vim.inspect(vim.b.zk))`
3. Confirm the buffer has zk info

#### Save a buffer (BufWritePost):
1. Modify title or heading
2. Save the buffer
3. Then `:lua print(vim.inspect(vim.b.zk))`
4. Confirm the title is changed

#### 1. autocmds.lua

I added `lua/zk/autocmds.lua` for this PR and for possible future additions.
If this doesn't feel right for the project at the moment , I'm happy to move the autocmd section into an existing file instead.

#### 2. slow down

Since it calls `zk.api.list` for each buffer, it may cause slow down when many buffers are opened in nvim start.
It can be fixed. But I should leave it as it is, for the cases that some buffers are opened from different notebook_path(s).

### DIFF
--- a/lua/zk/autocmds.lua
+++ b/lua/zk/autocmds.lua
@@ -5,8 +5,11 @@ local util = require("zk.util")
 ---Update zk cache on save and read
 vim.api.nvim_create_autocmd({ "BufWritePost", "BufReadPost" }, {
   pattern = { "*.md" },
-  callback = function()
+  callback = function(ev)
     local bufnr = vim.api.nvim_get_current_buf()
+    if ev.event == "BufReadPost" and vim.b[bufnr].zk then
+      return
+    end
     vim.schedule(function()
       util.zk_buf_cache(bufnr)
     end)


riodelphino:
Closed by [this comment](https://github.com/zk-org/zk-nvim/pull/260#issuecomment-3755511003)

PR #277, opened 2026-01-02
Make fzf-lua show_note_picker() take multi_select into account.

winter-again:
Currently, the `show_note_picker()` function for the fzf-lua picker doesn't take into account the value of `multi_select`.


winter-again:
I assumed that this change was too small to warrant creating a separate branch for the PR. Let me know if it should be on its own branch and I can close this PR and open a new one.

tjex:
But, I can multi select already without this change.. You can't?

```
 24 - a goal or a function           │Zk Notes ❯                                                                            520/520 (2) │
 23                                  │  CTRL-E: create a note with the query as title                                                   │
 22 Zettelkasten embodies systems thi│ |...because they build connections                                                               │
 21                                  │ |A Definition of Systems Thinking: A Systems Approach                                            │
 20 ~ p.3                            │▌ A New Method of Making Common-Place-Books                                                       │
```

PR #278, opened 2026-01-02
Fix get_buffer_paths() function used by ZkBuffers command so that only active buffers are listed

winter-again:
The buffers returned by `ZkBuffers` include inactive/deleted buffers despite the docs/annotations stating that the command is only for active buffers. This adds a check for valid buffers using the suggested `vim.api.nvim_buf_is_loaded()` function.


tjex:
Nice one, thanks :)

PR #284, opened 2026-02-22
style: use opts with lazy.nvim

sumit-sahrawat:
This PR updates README.md to use `opts` table instead of a `config` function as [recommended by lazy.nvim](https://lazy.folke.io/developers#best-practices).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,10 @@ Via [lazy.nvim](https://github.com/folke/lazy.nvim)
 ```lua
 {
   "zk-org/zk-nvim",
-  config = function()
-    require("zk").setup({
-      -- See Setup section below
-    })
-  end
+  name = "zk",
+  opts = {
+    -- See Setup section below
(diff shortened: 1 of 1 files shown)


sumit-sahrawat:
@tjex The changes in this PR work for me. I've tested them.

For your sample config, please try with `name = "zk"`. The `name` property is used to decide what's passed to `require` by `lazy.nvim` in the `require(<module>).setup(opts)` call. `require("zk-local")` is not a thing so it doesn't work.

Ideally the repo would be named `zk.nvim` and lazy would automatically use `require("zk")` to call setup but that might be overkill.

I've updated the troubleshooting section as well and grepped and found no other instances of `config = function`.

sumit-sahrawat:
@tjex we do need the `name` key. I'm getting the same `Lua module not found for config` error for both github and local repo if I don't set `name`.

PR #290, opened 2026-05-06
Remove "note-count" from `ZkTags` sort options before passing to notes picker

winter-again:
I found that using `ZkTags {sort = {"note-count"}}` gives an error once a tag is selected because the arguments in `options.sort` are immediately passed to `zk.edit`. However, the `"note-count"` argument can't be used to sort notes. According to [zk docs](https://zk-org.github.io/zk/tips/editors-integration.html#zk-tag-list), `name` and `note-count` are the only allowed arguments, so I considered it safe to simply remove `note-count`.

More broadly, I think this is related to what I consider a shortcoming in using `ZkTags` to search for notes. From my understanding, there's no way to explicitly pass separate tables of search options (i.e., one for the tags picker and one for the subsequent notes picker). A while ago I was playing around with a custom command that internally handles this; nothing fancy, just pulling out `note-count` similar to what I've done here. Not sure if you consider such functionality useful.

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -81,7 +81,7 @@ local function insert_link(selected, opts)
   opts = vim.tbl_extend("force", {}, opts or {})
 
   local location = util.get_lsp_location_from_selection()
-	local selected_text = ""
+  local selected_text = ""
 
   if not selected then
     location = util.get_lsp_location_from_caret()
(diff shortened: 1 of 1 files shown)


tjex:
Ooh, this is a good catch. Thanks. Sorting by name was also broken, as name is also not a valid sort value for the subsequent zk list sort input. I adapted your solution to take a table of sort options, does this still fix the issue on your end?

winter-again:
Yes, works for me. Thanks.

PR #292, opened 2026-05-18
Only filter ZkTags sort options if present

winter-again:
#290 introduced a bug with `ZkTags` because the `options.sort` filtering was happening regardless of whether `options` or `options.sort` given. I must have only tested out the result when passing sort options to `ZkTags` and not a bare `ZkTags` call.


tjex:
Ups. Thanks!

PR #293, opened 2026-05-18
Add ability to combine target tags with OR

winter-again:
I realized that you can combine tags during a `ZkTags` search with an `"OR"` conjunction. The implicit default behavior appears to be `"AND"`. What do you think of introducing this option `options.use_or = true` (boolean) to the `ZkTags` command? Maybe this risks conflicts with other commands that don't use such an option (i.e., like the sort options issue in #290)? Or, maybe this needs to be done in parallel with changes in `zk`, not just `zk-nvim`?

This [issue](https://github.com/zk-org/zk/issues/722) has a similar sentiment.

### DIFF
--- a/lua/zk/commands/builtin.lua
+++ b/lua/zk/commands/builtin.lua
@@ -139,6 +139,10 @@ commands.add("ZkTags", function(options)
       end, options.sort)
     end
 
+    if options and options.use_or then
+      tags = { table.concat(tags, " OR ") }
+    end
+
     options = vim.tbl_extend("keep", { tags = tags }, options or {})
     zk.edit(options, { title = "Zk Notes for tag(s) " .. vim.inspect(tags) })
   end)


tjex:
Sorry for the late reply. Yes I would want to address this "at the source" (zk), and keep zk-nvim as a neovim interface for zk. 

If we start building in functionality that users see as a 'zk functionality', but it doesn't work in zk cli, this gets confusing. Thanks all the same.

winter-again:
Makes sense. Thanks.

tjex:
@winter-again I got a little confused. It's already possible to list tags with 'OR' in zk. How are you trying to achieve this in zk-nvim, but without success?

As in, what options are you providing to `ZkTags`

Actually, how are you providing anything to `ZkTags` other than `{sort = {<sort-option}}`?

Aren't you meaning `ZkNotes {tags = {'tag1', 'tag2'}}`?

In that case you can pass `ZkNotes {tags = 'tag1 OR tag2'}}` to combine target tags in a search.

winter-again:
@tjex I think my issue was related to calling `ZkTags`, selecting multiple tags, and wanting the resulting notes picker to join the selected tags using OR instead of implicit AND. Please correct me if that's already possible. I may have missed it. Edit: your suggestion is to simply use `ZkNotes {tags = ...}` instead of passing through a tags picker, right?

My solution was to write a custom command that joined selected tags with OR before passing them on to the notes picker. When I looked back at the code for this PR, I thought it might be too simplistic or maybe not the right level to make such a change (i.e., your earlier comment about wanting to make such changes at `zk` level rather than nvim plugin).

tjex:
Ahhh right. So multiselecting tags in the `ZkTags` picker to return notes tagged `<selection 1> OR <selection n>`.

Then that would be something to implement on the zk-nvim side. 
But then I would say an option `options.tags.multiselect_strategy = <OR/AND>' would be the way to go, and have it default to `AND`. Instead of switching `OR` on or off.

This should be overwritable by passing an option in the command execution though. So if the strategy is set to `OR` in the config, I should be able to pass `ZkTags {multiselect_strategy = 'OR'}`.

I can't reopen because you deleted the branch. You can just make a new PR if you like and link to this to document the context.

winter-again:
Sure. I can put something together later this week. Thanks.

PR #294, opened 2026-06-13
Add multi_select_strategy option for configuring how multiple tags are treated during search

winter-again:
See #293 for origin/context.

- Adds plugin config `tags.multi_select_strategy` option that defaults to `"AND"`
- Uses strategy given by options when `ZkTags` used. Otherwise, falls back to plugin config.
- Instead of specifying tags as a table of tags, I standardized by always converting the table of tags to a "query" string using the specified separator (`"AND"` or `"OR"`).
- Also display the chosen separator in the notes picker prompt
- Adds option to default `lazy.nvim` config example in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ return {
         enabled = true,
       },
     },
+
+    tags = {
+      -- Configure how multiple tags should be combined in a ZkTags search
+      -- Can be "AND" or "OR"
+      multi_select_strategy = "AND",
+    }
   },
 }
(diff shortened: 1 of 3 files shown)


tjex:
Sorry again for dealys and thanks for resubmitting. Looks good and all working on my end.

tjex (approved):

winter-again:
No problem. Thanks for maintaining `zk`.

PR #298, opened 2026-08-24
Extend ZkMatch to allow free text and text under cursor search functionality.

tjex:
In pursuit of better grepping / searching like #171 #232. Still need zk --match to ideally return word positions to allow highlighting search term position results.

Currently, ZkMatch operates only on visually selected text. This PR enables the
following extensions to the function:

Search for notes containing "my" "search" & "term" (all words anywhere in the file):
`:ZkMatch my search term`

Search for notes containing "my search term" (exact string):
`:ZkMatch "my search term"`

Cursor on word "term":
`:ZkMatch <cursor-under-word. i.e., 'term'>`

Cursor visually selecting "my search term" (existing behaviour, all words anywhere in the note):
`:'<,'>ZkMatch my search term`

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,15 @@ see what they can do, and learn as you go.
 
 [Options (ZkList)](https://zk-org.github.io/zk/tips/editors-integration.html#zk-list)
 
-- `:'<,'>ZkMatch [{options}]`\
-  Opens a notes picker, filtering for notes matching the visual selection.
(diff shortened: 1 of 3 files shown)


tjex:
@riodelphino Hopefully this is a partial solution to at least be able to search through notes with titles being returned in your picker. As said, we still need to get around to looking at getting range positions returned by zk --match.

PR #301, opened 2026-08-28
Fix ZkIndex crash from unsolicited range argument

### LINKED ISSUE #300: `zkIndex` command fails with "attempt to call upvalue 'cb' (a number value)" — command wrapper passes range as callback
---

### Description

Running `:ZkIndex` (no visual selection) throws:

```
vim.schedule callback: .../zk-nvim/lua/zk.lua:89: attempt to call upvalue 'cb' (a number value)
stack traceback:
	.../zk-nvim/lua/zk.lua:89: in function 'handler'
	.../nvim/runtime/lua/vim/lsp/client.lua:743: in function 'fn'
	[string "vim/_core/editor"]:273: in function <[string "vim/_core/editor"]:272>
```

Likely root cause is that `commands/builtin.lua` registers the command directly against the library function:

```lua
commands.add("ZkIndex", zk.index)
```

`commands/init.lua`'s generic wrapper invokes every registered command as:

```lua
fn(loadstring("return " .. params.args)(), params.range)
```

This unconditionally passes `params.range` as the **second** positional argument to whatever `fn` is. For `ZkIndex`, that second argument lands in `zk.index(options, cb)`'s `cb` parameter. With no selection, `params.range` is `0`.

Because Lua only treats `nil`/`false` as falsy (unlike JS/Python, where `0` is falsy), the fallback in `zk.lua`:

```lua
function M.index(options, cb)
  options = options or {}
  cb = cb or function(stats)
    vim.notify(vim.inspect(stats))
  end
  ...
```

never fires — `cb` stays `0`, a number. When the LSP response for `workspace/executeCommand` arrives asynchronously, `cb(stats)` at `zk.lua:89` throws because `0` isn't callable.

### Reproduction (headless, no `zk` binary required)

```lua
-- repro.lua
package.path = "<path-to>/zk-nvim/lua/?.lua;<path-to>/zk-nvim/lua/?/init.lua;" .. package.path

local api = require("zk.api")
api.index = function(path, options, cb)
  vim.schedule(function()
    cb(nil, { updated = 1, added = 0 })
  end)
end

local commands = require("zk.commands")
local zk = require("zk")
commands.add("ZkIndex", zk.index)

vim.cmd("ZkIndex")
vim.defer_fn(function() vim.cmd("qa") end, 200)
```

```
nvim --headless -u NONE -l repro.lua
```

### Environment

- Neovim 0.12.5
- zk-nvim: current `main`

I can put together a fix. Not sure of the contribution rules.

tkolleh:
Fixes #300 — running `:ZkIndex` with no visual selection crashes with:

```
vim.schedule callback: .../zk-nvim/lua/zk.lua:89: attempt to call upvalue 'cb' (a number value)
```

### Root cause

`lua/zk/commands/init.lua`'s generic dispatcher (`commands.add`) called every registered command as `fn(loadstring("return " .. params.args)(), params.range)` — unconditionally forwarding `params.range` as the second positional argument, regardless of what the handler function actually expects there.

`lua/zk/commands/builtin.lua` registers `commands.add("ZkIndex", zk.index)`, passing the library function directly. `zk.index(options, cb)` expects that second argument to be a callback. With no selection, `params.range` is `0`. Lua only treats `nil`/`false` as falsy (unlike JS/Python, where `0` is falsy), so the fallback `cb = cb or function(stats) ... end` in `zk.lua` never fires — `cb` stays the number `0`, and `cb(stats)` throws once the async LSP response for `workspace/executeCommand` arrives.

## Fix

I checked every other `commands.add(...)` registration in `builtin.lua` for the same latent pattern. `ZkCd` also registers a bare library function (`zk.cd`) instead of an inline wrapper, but `zk.cd(options)` only declares one parameter, so Lua silently discards the extra argument — no crash. `ZkMatch` is the only other command that receives `range` today, and it already opts in via `optional_selection` and declares `range` as a named parameter, so it's unaffected by this change. `ZkIndex` was the only command with the bug.

Rather than special-casing `ZkIndex` in `builtin.lua`, this fixes the dispatcher itself: `params.range` is now only forwarded to a command's handler if that command was registered with `needs_selection` or `optional_selection`. Commands that don't opt in are called with a single `options` argument, matching what every command handler in `builtin.lua` already expects except for the two bare-function registrations.


tjex (approved):
Excellent, thank you :)

